Issue #78, opened 2014-06-05
SAML2:EncryptedAssertion support


gheeres:
Interfacing with [Shibboleth IDP](https://shibboleth.net/).

```xml
<saml2:EncryptedAssertion xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion">
</saml2:EncryptedAssertion>
```

explunit:
I see there has been some progress toward supporting encrypted assertions. Is there anything I can do to help with this effort? E.g. It looks like #85 from @gheeres was close to being merge-able. Is there value in me picking up that code and modifying it to address the comments from @AndersAbel?

Or do we think the branch from @Raschmann is going to be moving toward a PR sometime soon?

PoyaManouchehri:
We also really need this feature. Would be happy to help with anything to get that branch merged.

AndersAbel:
I think that #85 is quite close to being possible to merge. Next step would be to pick up the discussion and answer my comment from June 11th 2014. As a start, please feel free to comment.

Another option is that you sponsor the development of the feature. Please mail me if that is of interest.

PoyaManouchehri:
I did comment on that PR. Any thoughts?

AndersAbel:
@PoyaManouchehri Sorry for not getting back to you. I've been extremely busy the last few weeks. I'll get back to you as soon as I have time.

rdefreitas:
Looks like it still had a few merge conflicts. Do you need help getting this resolved?

AndersAbel:
@rdefreitas Help getting this done would be appreciated.

PR #85, opened 2014-06-09
78 encrypted assertion

gheeres:
Implementation of issue #78.

Summary
-------------
+ Refactored certification reading into helper class. 
+ Created custom exceptions for certificate handling. 
+ Created helper / static class to decrypt an EncryptedAssertion replacing the contents of the Saml2Response with the decrypted contents.

Questions
-------------
Due to the dependency on the Windows certificate store, I'm unsure how to proceed with writing unit tests without creating an unnecessary abstraction. The previous CertificateElement class didn't handle this code coverage either... are you OK with this?

Also unsure on how to proceed with testing of message decryption. If full testing is desired, then we need a test helper able to generate EncryptedAssertion messages. Does anything exist yet?

### DIFF
--- a/Kentor.AuthServices.Tests/CertificateNotFoundExceptionTests.cs
+++ b/Kentor.AuthServices.Tests/CertificateNotFoundExceptionTests.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using FluentAssertions;
+
+namespace Kentor.AuthServices.Tests
+{
+    [TestClass]
+    public class CertificateNotFoundExceptionTests
+    {
+        [TestMethod]
+        public void CertificateNotFoundException_DefaultCtor()
+        {
+            ExceptionTestHelpers.TestDefaultCtor<CertificateNotFoundException>();
+        }
+
+        [TestMethod]
+        public void CertificateNotFoundException_SerializationCtor()
+        {
+            ExceptionTestHelpers.TestSerializationCtor<CertificateNotFoundException>();
+        }
+
+        [TestMethod]
+        public void CertificateNotFoundException_StringCtor()
+        {
+            var msg = "Message!";
+            var subject = new CertificateNotFoundException(msg);
+
+            subject.Message.Should().Be(msg);
+        }
+
+        [TestMethod]
+        public void CertificateNotFoundException_InnerExceptionCtor()
+        {
+            var inner = new Exception("inner");
+            var msg = "Message!";
+
+            var subject = new CertificateNotFoundException(msg, inner);
+
+            subject.Message.Should().Be(msg);
+            subject.InnerException.Should().Be(inner);
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


PoyaManouchehri:
@AndersAbel and @gheeres : Just some feedback regarding the issue of replacing the EncryptedAssertion XML node with an Assertion note:

I think the essence of the issue is that `Saml2Response` class has probably grown and is doing a bit too much now. Ideally there would be something like `Saml2ResponseXml` which manages the reading and writing, and then `Saml2Response` would be a logical representation that deals with Claims and metadata, and not directly with XML. Then there would then be some mapping code between the two that also deals with the encryption/decryption. There would be no need to replace the XML nodes in this case.

As a side note this code only deals with decryption, not encryption of a SAML assertion (which is fine for most use cases of this library I would imagine...)

Anyway the change I'm suggesting is probably a bigger one and outside of scope for this. I'm happy with this pull request as is (and would prefer it goes in sooner rather than later), but maybe we can create an issue for a refactor. I would be happy to contribute to that.

AndersAbel:
@PoyaManouchehri Yes, the `Saml2Response` class does to much. One thing I'd like to move out is the XML signature stuff - and also the decryption when encrypted assertions are supported.

The reading of XML could be moved to the `SAML2PSecurityTokenHandler` class to follow the convention by the rest of `System.IdentityModel`. We just have to make sure that the `Saml2Response` class stays immutable if that is done.

I agree that those steps can be made as a separate refactoring. To get this PR merged though, I'd prefer to not replace the XML nodes when decrypting the data, but instead add them to the `allAssertionElementNodes` collection. That should be quite a small change compared to the existing PR.

PoyaManouchehri:
Just to follow up, is anyone looking at modifying the PR? @gheeres or @AndersAbel? If not I can look into making the changes

AndersAbel:
@PoyaManouchehri I've not had time to look at it, and I haven't heard from @gheeres for long. If you want to bring this up to date it would be appreciated.

gheeres:
Sorry, I haven't been following this. I'm no longer using this project. We've changed our AUTH strategy to use [ADFS](https://msdn.microsoft.com/en-us/library/bb897402.aspx) which .NET WIF natively supports. ADFS is able to communicate with our SAML2 infrastructure ([Shibboleth](https://shibboleth.net/)).

AndersAbel:
@gheeres No problem, I completely understand and am grateful for the work you've done, which I hope we'll be able to take forward and merge.

explunit:
I made [some progress](https://github.com/explunit/authservices/tree/encrypted-assertions) on bring this PR up to date but in the end I decided that the certificate lookup code in this branch was not at the right place. The SP certificate lookup should be part of the general configuration & metadata and eventually shared with #95.

Also, the decryption approach in this branch breaks signature validation when the overall response is signed (but works OK when signature was part of the encrypted assertion itself). The spec (section 6.2) is a little vague on this point but that's how I read it.

Of course I'm willing to be corrected on any of this if I'm mistaken. I could continue this branch if we feel it's better than #273

AndersAbel:
Closing as this is implemented by @explunit in #273.

PR #155, opened 2014-11-06
Session token lifetime configuration

Raschmann:
Session token lifetime configuration based on `<sessionTokenRequirement>` (if present) `lifetime` attribute.


AndersAbel:
Looked into this and won't merge this. The lifetime should take SessionNotOnOrAfter into account and the solution should work with all three API modules (owin, HttpModule, MVC)

PR #168, opened 2014-11-27
AuthnRequest signing support in HTTP Redirect binding

Raschmann:
AuthnRequest signing implementation and tests


AndersAbel:
Closing due to no lack of activity. I'll open it again if there is any more done.

Issue #172, opened 2014-12-08
Validate federation metadata signature


AndersAbel:
AuthServices should validate the signature of the metadata supplied by federations. It should be possible to add federation certificates to configuration and make authservices only accept metadata that is signed by those certificates.

AndersAbel:
Changing this issue to only concern federations and making a separate issue for identity providers.

PR #187, opened 2015-01-27
Proposed fix to #172: Validate metadata

TobbeHolmstrom:
Added metadata signature validation with a specified certificate file or with other methods. StubIdp now adds signatures to the metadata.

Test coverage is not 100%. The known remaining work is to get the remaining 2
ignored tests to work.


AndersAbel:
Metadata signature validation has been implemented.

Issue #190, opened 2015-02-11
Metadata download - multiple signing keys (key rollover)


doormalena:
Hi!

With 0.12.0 (and below) the following exception occurs when the IDP does a rollover of the signing certificate and the SP (Kentor) is configured to auto-download the metadata at startup. During a rollover there are two signing certificates for a period of time available in the metadata. Depending on where we are in time within this period, either the first or the second is valid. The IDP product we use is ADFS 2012R2.

If you want I can send you a fix for this. I have changed Saml2Response and IdentityProvider classes to support multiple signing certificates.

-------------------

Sequence contains more than one element 		
			at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source)
   at Kentor.AuthServices.IdentityProvider.ReadMetadataIdpDescriptor(ExtendedEntityDescriptor metadata) in d:\Data\Workspaces\doormalena\authservices-0.12.0\Kentor.AuthServices\IdentityProvider.cs:line 328
   at Kentor.AuthServices.IdentityProvider.ReadMetadata(ExtendedEntityDescriptor metadata) in d:\Data\Workspaces\doormalena\authservices-0.12.0\Kentor.AuthServices\IdentityProvider.cs:line 307
   at Kentor.AuthServices.IdentityProvider.DoLoadMetadata() in d:\Data\Workspaces\doormalena\authservices-0.12.0\Kentor.AuthServices\IdentityProvider.cs:line 275
   at Kentor.AuthServices.IdentityProvider.set_LoadMetadata(Boolean value) in d:\Data\Workspaces\doormalena\authservices-0.12.0\Kentor.AuthServices\IdentityProvider.cs:line 108
   at Kentor.AuthServices.IdentityProvider..ctor(IdentityProviderElement config, ISPOptions spOptions) in d:\Data\Workspaces\doormalena\authservices-0.12.0\Kentor.AuthServices\IdentityProvider.cs:line 46
   at Kentor.AuthServices.Configuration.IdentityProviderCollection.RegisterIdentityProviders(IOptions options) in d:\Data\Workspaces\doormalena\authservices-0.12.0\Kentor.AuthServices\Configuration\IdentityProviderCollection.cs:line 58
   at Kentor.AuthServices.Configuration.Options.get_FromConfiguration() in d:\Data\Workspaces\doormalena\authservices-0.12.0\Kentor.AuthServices\Configuration\Options.cs:line 25
[...]

Issue #105, opened 2014-09-02
Support for multiple keys per idp


AndersAbel:
An idp can have multiple keys. It should be possible both to configure manually and to load multiple keys through metadata. This could e.g. be used when the idp changes keys.

Issue #98, opened 2014-08-01
ValidateInResponseTo results in "The Saml2Response didn't pass validation" in a farm setup


doormalena:
ValidateInResponseTo checks if the responseId was issued during a previous AuthN request. This responseid is cached in a static variable within PendingAuthnRequests. This variable however is different per node in a web farm. When a load balancer decides to route the returning request to a different node, the ValidateInResponseTo method is unable to find the responseId in it's PendingAuthnRequests cache. This results in  "The Saml2Response didn't pass validation" while the response is perfectly valid.

martinjt:
@AndersAbel Could we simply create an interface to the PendingAuthnRequests, then use that in the code.  Then additionally, allow it to be overridden when the IdentityProvider object is created.  That would be a great start, and would allow people to create their own implementations until you've decided on yours.  Beyond that, it would allow a simple way to support both the old (application state) and new methods (cookie/redis)?

I can do that implementation and submit a pull request.  We want to use the library as is, without modification, so I'm happy to invest the time if you can give me some assurance that you will be able to review and merge within the next few weeks?  It should be a relatively minor change to just add the interface and make it public so it can be set, then set the default to the existing implementation?

Sidenote: using cookies would rely on the site supporting cookies, although it's not a problem for us, some folks might not like using them.  Are there any other dependencies on cookies that would make that point m00t?  I'm no security expert, but SAML authentication seems to be something that, unless done right, could be catastrophic, so allowing the user to choose the implementation of storing the replay stuff would be beneficial.

martinjt:
@AndersAbel I've put together an outline of my thoughts on the interface in this commit:

https://github.com/martinjt/authservices/commit/eb332c5400c0a948c29da2ef27cccde745e4dff5

it needs cleaning up, but I think it provides a decent extensibility point.

AndersAbel:
Adding an interface as the extensibility point is a good start. I don't think I'll have time to do the cookie-based solution for quite some time anyway.

When the cookie-based solution discussed above is completed, it can be added as another provider. It has some drawbacks, as the amount of data that need to be serialized can be quite large.

I briefly looked through the code and it looks like the right approach.

akiander:
[...]

Issue #182, opened 2015-01-15
Owin Middleware doesn't preserve AuthenticationProperties across Discovery Service


AndersAbel:
Similar to #127, but that is about preserving the properties when doing an AuthnRequest and processing the response. To have a complete solution it needs to be preserved across Discovery Services calls too.

federicobarera:
Hi @AndersAbel ,

I came across authservices while investigating Saml2p support in identity server. 
Could you please spare a minute to extend on the implications of these limitation? Which flows/ bindings affects? Thanks

Federico

AndersAbel:
This particular issue relates to when there is a Discovery Service configured with AuthServices. That is typically the case when you have a federation with a dynamically loaded list of identity providers. The Discovery Services is a UI for letting the user chose hat identity provider to use for authentication.

Using a discovery service is an advanced scenario. If you haven't configured one and nobody on the identity provider side has ever mentioned the need of one, you're not affected.

wagich:
I debugged the issue and made a few changes to persist the state of the AuthenticationProperties using the existing StoredRequestState facilities. The solution works beautifully but I'd quite like a second opinion on my implementation. If all looks good I'd be happy to make a pull request.

You can have a look at the code on https://github.com/wagich/authservices/tree/preserve-state-across-ds

AndersAbel:
@wagich Looks interesting. Please submit a PR to get it through the automatic bild and code coverage.

PR #194, opened 2015-02-13
Support for multiple signing certificates in downloaded metadata (#105) + Fix for ValidateInResponseTo results in "The Saml2Response didn't pass validation" in a farm setup #98

doormalena:
Exisiting code is altered to fix issue reported in #190.  
- Kentor.AuthServices/IdentityProvider.cs
- Kentor.AuthServices/SAML2P/Saml2Response.cs

Fix for ValidateInResponseTo results in "The Saml2Response didn't pass validation" in a farm setup #98
- Kentor.AuthServices/Internal/PendingAuthnRequests.cs

### DIFF
--- a/Kentor.AuthServices/IdentityProvider.cs
+++ b/Kentor.AuthServices/IdentityProvider.cs
@@ -77,7 +77,7 @@ namespace Kentor.AuthServices
                 throw new ConfigurationErrorsException("Missing binding configuration on Idp " + EntityId.Id + ".");
             }
 
-            if (SigningKeys == null)
+            if (!SigningKeys.Any())
             {
                 throw new ConfigurationErrorsException("Missing signing certificate configuration on Idp " + EntityId.Id + ".");
             }
--- a/Kentor.AuthServices/IdentityProvider.cs
+++ b/Kentor.AuthServices/IdentityProvider.cs
@@ -1,4 +1,5 @@
-﻿using Kentor.AuthServices.Configuration;
+﻿using System.Collections.Generic;
+using Kentor.AuthServices.Configuration;
 using System;
 using System.Configuration;
 using System.Globalization;
@@ -46,11 +47,11 @@ namespace Kentor.AuthServices
             LoadMetadata = config.LoadMetadata;
             this.spOptions = spOptions;
 
-            var certificate = config.SigningCertificate.LoadCertificate();
+            var certificate = config.SigningCertificate.LoadCertificate(); // TODO: add support for multiple configured signing certificates
 
             if (certificate != null)
             {
-                signingKey = certificate.PublicKey.Key;
+                signingKeys.Add(certificate.PublicKey.Key);
             }
 
             try
@@ -76,7 +77,7 @@ namespace Kentor.AuthServices
                 throw new ConfigurationErrorsException("Missing binding configuration on Idp " + EntityId.Id + ".");
             }
 
-            if (SigningKey == null)
+            if (SigningKeys == null)
             {
                 throw new ConfigurationErrorsException("Missing signing certificate configuration on Idp " + EntityId.Id + ".");
             }
@@ -243,23 +244,18 @@ namespace Kentor.AuthServices
             return Saml2Binding.Get(Binding).Bind(request);
         }
 
-        private AsymmetricAlgorithm signingKey;
+      private List<AsymmetricAlgorithm> signingKeys = new List<AsymmetricAlgorithm>();
 
         /// <summary>
         /// The public key of the idp that is used to verify signatures of responses/assertions.
         /// </summary>
-        public AsymmetricAlgorithm SigningKey
+        public IList<AsymmetricAlgorithm> SigningKeys
         {
             get
             {
                 ReloadMetadataIfRequired();
-                return signingKey;
+                return signingKeys;
             }
-            set
-            {
-                signingKey = value;
-            }
-
         }
 
         object metadataLoadLock = new object();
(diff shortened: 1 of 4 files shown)


albinsunnanbo:
Is the ReturnUrl support removed?

doormalena:
Thanks for the comment. I will have a look at it on monday. It was not intentionally removed.


-------- Original message --------
From: Albin Sunnanbo
Date:14/02/2015 08:41 (GMT+01:00)
To: KentorIT/authservices
Cc: Alexander van Doormalen
Subject: Re: [authservices] Support for multiple signing certificates in downloaded metadata (#105) + Fix for ValidateInResponseTo results in "The Saml2Response didn't pass validation" in a farm setup #98 (#194)


Is the ReturnUrl support removed?

Kentor.AuthServices/Internal/PendingAuthnRequests.cs
```diff
@@ -1,40 +1,94 @@
...
+        var cookie = HttpContext.Current.Request.Cookies[CookieName];
+
+        if (cookie != null)
+        {
+          var cookieId = Decrypt(cookie.Values[CookieIdAttribute]);
+          var cookieIdpId = Decrypt(cookie.Values[CookieIdpIdAttribute]);
+
+          if (id != null && cookieId == id.Value)
+          {
+            idp = new StoredRequestState(new EntityId(cookieIdpId), null); // TODO: returnUrl is left out here as it's not used. Possibly add it by also storing it into the cookie (or get it from config based in the id) or simply remove if from the stored state
```
albinsunnanbo: Here ReturnUrl seems to be lost?

Kentor.AuthServices/Internal/PendingAuthnRequests.cs
```diff
@@ -1,40 +1,94 @@
...
+      private const string CookieName = "AuthNRequest";
+      private const string CookieIdAttribute = "id";
+      private const string CookieIdpIdAttribute = "idpId";
+      private const string Purpose = "Initiator Verification";
+
+      internal static void Add(Saml2Id id, StoredRequestState idp)
+      {
+        var cookie = GetNewCookie();
+        cookie.Values[CookieIdAttribute] = Encrypt(id.Value);
+        cookie.Values[CookieIdpIdAttribute] = Encrypt(idp.Idp.Id);
```
albinsunnanbo: Here ReturnUrl seems to be lost?
RelayData too.
Serialization of the whole StoredRequestState object might be a better solution?

doormalena: Yes seems like a good proposal, also to support any additions/changes being done to the StoredRequestState. Thnx.

Kentor.AuthServices/Internal/PendingAuthnRequests.cs
```diff
@@ -1,40 +1,94 @@
...
+        {
+          try
+          {
+            var protectedBytes = Convert.FromBase64String(value);
+            var unprotectedBytes = MachineKey.Unprotect(protectedBytes, Purpose);
+            returnValue = unprotectedBytes != null ? Encoding.UTF8.GetString(unprotectedBytes) : String.Empty;
+          }
+          catch (Exception)
+          {
+            // TODO: logging / exception. Probable causes: tempering with cookie, or machine keys are not equal on each farm member
```
albinsunnanbo: TODO...

Kentor.AuthServices/IdentityProvider.cs
```diff
@@ -76,7 +77,7 @@ private void Validate()
                 throw new ConfigurationErrorsException("Missing binding configuration on Idp " + EntityId.Id + ".");
             }
 
-            if (SigningKey == null)
+            if (SigningKeys == null)
```
albinsunnanbo: I suppose !SigningKeys.Any() should also throw configuration error?

doormalena: As it's now a list, null check indeed became invalid. Needs to be changed to SigningKeys.Count == 0
Will do this on Monday as well.

Kentor.AuthServices/SAML2P/Saml2Response.cs
```diff
@@ -1,480 +1,487 @@
...
+                if (!allowedTransforms.Contains(transform.Algorithm))
+                {
+                    throw new Saml2ResponseFailedValidationException(
+                        "Transform \"" + transform.Algorithm + "\" found in Xml signature is not allowed in SAML.");
+                }
+            }
+
+            var validSignature = false;
+
+            for (var i = 0; !validSignature && i < idpKeys.Count; i++)
```
albinsunnanbo: or

    var validSignature = idpKeys.Any(key => signedXml.CheckSignature(key));

or even

    var validSignature = idpKeys.Any(signedXml.CheckSignature);

makes it more compact and readable

Kentor.AuthServices/Internal/PendingAuthnRequests.cs
doormalena: What's best practice for this in the AuthServices Library? Throw exception?

AndersAbel:
Thanks to both of you for a good discussion on the code in the Pull Request. Unfortunately, I have something that will cause some extra work to say: I want this PR to be split in two.

* One PR for the support for multiple keys (#105)
* One PR for the farm setup (#98 and #182)

Regarding the latter, I have been thinking quite a bit about how to approach the farm setup issue. It is a complex area, especially as the `StoredRequestState` provides protection against both replay attacks and XSRF attacks.

albinsunnanbo:
Hmmm, can replay attacks be prevented by an encrypted cookie? If you replay the cookie too the server will happily decrypt it and find it valid.
The replay state needs to be stored on the server side.
I can see several strategies for sharing state in a web farm.

0) Current local dictionary for simple setup in non farm solutions
1) Dedicated state server
2) Redis or some other key value store
3) Sync state between servers using the windows service bus

We might need some plugin strategy here.

doormalena:
If you want to prevent replay attacks, you also need to take into account the WIF created session. How is WIF protected against replay attacks? We can limit the replay window by adding a timestamp. 

If you don't want a cookie, we can also serialize, encrypt and base64 encode the StoredRequestState and send it as a RelayState parameter to the IDP. This however, doesn't prevent the replay, only prevents cookie re-usage.

A replay attack (after stealing the cookie) can only be prevent in this case by like we discussed earlier a server-side kept state (which can be a database, state server, shared cache mechanisme etc.). As discussed in #98, this makes the deployment a bit more complex.

Secondly, what about the AllowUnsolicitedAuthnResponse setting? If this is turned on, there is no verification at all.

doormalena:
@albinsunnanbo 
I have processed all your feedback. Due note that the cookie mechanisme does have it's downsides which need to be considered (other than possible replay). A cookie can only store 4k of information. Adding the relaystate to it could potentionally break this construction. Adding the relaystate to the IDP requests would be a better solution or a state based approach (DB, state server etc...).

AndersAbel:
Discussed this issue, and will close it as Alexander will create two separate pull requests for these features.

PR #197, opened 2015-02-27
SAML LogoutRequest added

grabmill:
Hello Anders,

One of my customers was using Kentor AuthServices, but now they required the use of SAML Logout. I added support for a SAML LogoutRequest, which should always be signed.

Some remarks:
- In the config I had to add a property to configure the Url where the LogoutRequest should be posted to.
- In the Mvc project in the AuthServicesController I added methods with the name LogOff. I wasn't entirely sure if I should create separate methods for this or reuse the existing SignOut. Right now, there is a redirect to SignOut after the Logout-cycle is complete.
- Because we need to know which user to sign out, we need to get his name somehow. There are several places where I could implement this, but in the end I decided to go for the HttpRequestBaseExtensions and get the NameIdentifier there from the ClaimsIdentity.

Kind regards,
Roger Frehen

### DIFF
--- a/Kentor.AuthServices.HttpModule/HttpRequestBaseExtensions.cs
+++ b/Kentor.AuthServices.HttpModule/HttpRequestBaseExtensions.cs
@@ -37,6 +37,7 @@ namespace Kentor.AuthServices.HttpModule
                 requestBase.ApplicationPath,
                 requestBase.Form.Cast<string>().Select((de, i) =>
                     new KeyValuePair<string, string[]>(de, ((string)requestBase.Form[i]).Split(','))),
+                HttpUtility.ParseQueryString(requestBase.Url.Query),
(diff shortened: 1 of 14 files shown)


AndersAbel:
Thanks a lot for the PR. I have a busy schedule for the next few weeks but will look at this as soon as possible.

barryofguilder:
@grabmill, I successfully used this pull request to implement SAML logout with my ADFS server!  You couldn't have done this at a better time since I was needing to implement this for a customer in a very short time frame.  Thanks!

AndersAbel:
The code looks good, but I won't merge this as is with two different fixes in the same PR. Please see http://coding.abel.nu/2015/03/fixing-a-pull-request-from-master/ for a guide on how to fix it.

Issue #137, opened 2014-10-16
Change Saml2PSecurityTokenHandler to use ValidateToken()


AndersAbel:
Change `Saml2PSecurityTokenHandler` so that it can use the full `Validate()` method of `Saml2SecurityTokenHandler`. Add a new override for `ReadToken()` that accepts an issuer token as parameter, that is set (probably by copying data to a new instance) on the returned `Saml2SecurityToken`. The issuer token should correspond to the signature on the Saml2Response.

AndersAbel:
It looks like the signed assertion support is broken. I've disabled two tests when working on #66. When doing this issue, those two tests should be fixed again.

* Saml2Response_GetClaims_CorrectSignedMultipleAssertionInResponseMessage
* Saml2Response_GetClaims_CorrectSignedSingleAssertionInResponseMessage

PR #204, opened 2015-03-12
Fix signed assertions

### LINKED ISSUE #191: Signed Assertion handling broken?
A couple of bug reports from external users suggest that the signed assertion verification is somehow broken. There are also a couple of tests for signed assertions that are currently disabled: `Saml2Response_GetClaims_CorrectSignedSingleAssertionInResponseMessage` and `Saml2Response_GetClaims_CorrectSignedMultipleAssertionInResponseMessage`. When improving the error handling in #66 I found a problem previously hidden by the very rough error handling.

explunit:
Fixes #191 and moves in the direction mentioned in #137

### DIFF
--- a/Kentor.AuthServices.TestHelpers/SignedXmlHelper.cs
+++ b/Kentor.AuthServices.TestHelpers/SignedXmlHelper.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.IO;
-using System.Text.RegularExpressions;
-using System.Security.Cryptography.X509Certificates;
+﻿using System.Security.Cryptography.X509Certificates;
 using System.Xml;
 using Kentor.AuthServices;
 using System.Security.Cryptography.Xml;
@@ -14,9 +11,6 @@ namespace Kentor.AuthServices.TestHelpers
     public class SignedXmlHelper
     {
         public static readonly X509Certificate2 TestCert = new X509Certificate2("Kentor.AuthServices.Tests.pfx");
-        public static readonly X509SigningCredentials SigningCredentitals
-            = new X509SigningCredentials(TestCert, 
-                SecurityAlgorithms.RsaSha1Signature, SecurityAlgorithms.Sha1Digest);
 
         public static readonly AsymmetricAlgorithm TestKey = TestCert.PublicKey.Key;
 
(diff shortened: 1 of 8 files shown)


Kentor.AuthServices/IdentityProvider.cs
```diff
@@ -262,6 +263,19 @@ public AsymmetricAlgorithm SigningKey
...
+        /// The idp certificate stored locally, if any
+        /// </summary>
+        public X509Certificate2 Certificate
+        {
+            get
+            {
+                return certificate;
+            }
+        }
+
```
explunit: I'm not happy with this part, but could not figure out another way to get the info needed by the TokenResolver. If there's a way to construct `SecurityToken` from just the `SigningKey`, let me know.

Kentor.AuthServices.TestHelpers/SignedXmlHelper.cs
```diff
@@ -21,7 +21,7 @@ public class SignedXmlHelper
 
         public static string SignXml(string xml)
         {
-            var xmlDoc = new XmlDocument { PreserveWhitespace = true };
+            var xmlDoc = new XmlDocument { PreserveWhitespace = false };
```
explunit: Even after fixing the main issue with TokenResolver, this change was needed to suppress a "Digest Validation Failed" error in `Saml2SecurityTokenHandler.ReadToken`. The whitespace from the test cases was confusing WIF somehow.

AndersAbel:
Thanks a lot for this (and your other PRs). I've been terribly busy the last few weeks and haven't had time to look at it. I'll review these as soon as possible, next week looks less busy.

explunit:
@AndersAbel I fixed the issue (and added test case) with Idp's that have only metadata Url and no local certificate. 

However, the problem remains with overwriting the IssuerTokenResolver on the shared configuration.
WIF ignores any issuerToken passed into the `readToken` method, and even within `readAssertion` it is using the value from Configuration, so creating our own version of `readToken` will not solve it:
http://referencesource.microsoft.com/#System.IdentityModel/System/IdentityModel/Tokens/Saml2SecurityTokenHandler.cs,2644

One solution might be to construct a new token handler during each request rather than using a shared instance, but I'm not sure of the performance impact.

explunit:
Closing this for now, as it will conflict with other changes that @AndersAbel has in progress

PR #210, opened 2015-03-13
Sha256 support

### LINKED ISSUE #36: SHA256 support
CheckSignature method on SignedXml is not able to work with RSA-SHA256 method. Please implement a solution as described at "http://geekswithblogs.net/mkoerner/archive/2013/07/12/saml2-federationmetadata-validation.aspx"

explunit:
Fixes #36

### DIFF
--- a/Kentor.AuthServices/SAML2P/Saml2Response.cs
+++ b/Kentor.AuthServices/SAML2P/Saml2Response.cs
@@ -369,6 +369,7 @@ namespace Kentor.AuthServices.Saml2P
         /// <summary>Checks the signature.</summary>
         /// <param name="signedRootElement">The signed root element.</param>
         /// <param name="idpKey">The assymetric key of the algorithm.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage( "Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "AuthServices" )]
         private static void CheckSignature(XmlElement signedRootElement, AsymmetricAlgorithm idpKey)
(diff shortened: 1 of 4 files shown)


Kentor.AuthServices/SAML2P/Saml2Response.cs
```diff
@@ -369,6 +369,7 @@ private void ValidateSignature(IOptions options)
         /// <summary>Checks the signature.</summary>
         /// <param name="signedRootElement">The signed root element.</param>
         /// <param name="idpKey">The assymetric key of the algorithm.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage( "Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "AuthServices" )]
```
AndersAbel: Better to add AuthServices to CustomDictionary.xml (I'm surprised it isn't already listed there).

explunit: Adding AuthServices to CustomDictionary.xml creates dozens of code analysis errors like "CA1702 Compound words should be cased correctly. The compound word 'AuthServices' in namespace name 'Kentor.AuthServices.Metadata' exists as a discrete term. If your usage is intended to be single word, case it as 'Authservices'. Kentor.AuthServices	(Global)"

AndersAbel: Had to add it in both as a recognized word and a DiscreteException. Fixed in 522ff01

PR #212, opened 2015-03-19
Implement constructor for supplying full Urls to AuthServicesUrls

### LINKED ISSUE #211: Support AssertionConsumerServices path other than "Acs"
AuthServicesUrls currently adds a hardcoded "Acs" for the AssertionConsumerServices:
https://github.com/KentorIT/authservices/blob/master/Kentor.AuthServices/WebSSO/AuthServicesUrls.cs#L66

If you use the request and response classes separately (not using the HttpModule or the Mvc classes) this becomes awkward.

One solution would be to add a constructor to `AuthServicesUrls` which accepts two Uri's -- one for `AssertionConsumerServiceUrl` and the other for `SignInUrl`

I can create this pull request if you think this is an OK change.

explunit:
Fixes #211 

I'm not 100% happy with having to cast a null in some cases to disambiguate the constructors. Let me know if you see a better way.

### DIFF
--- a/Kentor.AuthServices.Tests/WebSSO/AuthServicesUrlsTests.cs
+++ b/Kentor.AuthServices.Tests/WebSSO/AuthServicesUrlsTests.cs
@@ -38,7 +38,7 @@ namespace Kentor.AuthServices.Tests.WebSso
         [TestMethod]
         public void AuthServicesUrls_Ctor_NullCheckModulePath()
         {
-            Action a = () => new AuthServicesUrls(new Uri("http://localhost"), null);
+            Action a = () => new AuthServicesUrls(new Uri("http://localhost"), (string)null);
 
             a.ShouldThrow<ArgumentNullException>("modulePath");
         }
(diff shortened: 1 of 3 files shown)


Kentor.AuthServices.Tests/WebSSO/AuthServicesUrlsTests.cs
```diff
@@ -38,7 +38,7 @@ public void AuthServicesUrls_Ctor_NullCheckApplicationUrl()
         [TestMethod]
         public void AuthServicesUrls_Ctor_NullCheckModulePath()
         {
-            Action a = () => new AuthServicesUrls(new Uri("http://localhost"), null);
+            Action a = () => new AuthServicesUrls(new Uri("http://localhost"), (string)null);
```
AndersAbel: I agree that casting of null isn't elegant. It's better to use a named argument:

`new AuthServicesUrls(new Uri("http://localhost"), modulePath: null)`

Kentor.AuthServices.Tests/WebSSO/AuthServicesUrlsTests.cs
```diff
@@ -77,5 +77,35 @@ public void AuthServicesUrls_Ctor_ChecksModulePathStartsWithSlash()
...
+            var subject = new AuthServicesUrls(null, new Uri("http://localhost/signin"));
+
+            subject.AssertionConsumerServiceUrl.Should().Be(null);
+            subject.SignInUrl.ToString().Should().Be("http://localhost/signin");
+        }
+
+        [TestMethod]
+        public void AuthServicesUrls_Ctor_NullCheckSignin()
+        {
+            Action a = () => new AuthServicesUrls(new Uri("http://localhost/signin"), (Uri)null);
```
AndersAbel: Used a named argument here too instead of the cast.

Kentor.AuthServices/WebSSO/AuthServicesUrls.cs
```diff
@@ -39,6 +39,7 @@ public AuthServicesUrls(HttpRequestData request, ISPOptions spOptions)
         /// </summary>
         /// <param name="applicationUrl">The full Url to the root of the application.</param>
         /// <param name="modulePath">Path of module, starting with / and ending without.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage( "Microsoft.Design", "CA1057:StringUriOverloadsCallSystemUriOverloads" )]
```
AndersAbel: Looked this up to understand what it warned about. Adding a Justification to the SupressMessage will save others from getting curious (maybe "Incorrect warning. modulePath isn't a string representation of a Uri".)

AndersAbel:
It's okay as it is, but as you pointed out casting `null` isn't very elegant (even though it can be okay in the tests). If you prefer to update, then do that, or I can merge directly.

PR #216, opened 2015-03-31
Fix #215 - Include parsed SAMLResponse in exceptions

### LINKED ISSUE #215: Include parsed SAMLResponse in exceptions thrown from AcsCommand.Run
Are there any security implications of including the SAMLResponse in the exception?
I assume not since the SAMLResponse can be intercepted in the client browser anyway when it is redirected or posted to the Acs command.

Maybe if the exception show up in an unprotected log it could be used for replay attacks?

albinsunnanbo:
Include parsed SAMLResponse in exceptions thrown from AcsCommand.Run
Fixes #215 

I put together a solution where the unpacked Saml2Response-string is added to the Exception.Data dictionary.
Please have a look at the way I "monkey patch" the payload onto the exceptions in AcsCommand:58
I prefer to add the payload to the existing exception instead of wrapping in a new exception with the old as an inner exception to preserve the original Exception type and not hide the original exception to many onion layers down.

### DIFF
--- a/Kentor.AuthServices.Tests/WebSSO/AcsCommandTests.cs
+++ b/Kentor.AuthServices.Tests/WebSSO/AcsCommandTests.cs
@@ -86,7 +86,31 @@ namespace Kentor.AuthServices.Tests.WebSso
 
             a.ShouldThrow<BadFormatSamlResponseException>()
                 .WithMessage("The SAML response contains incorrect XML")
-                .WithInnerException<XmlException>();
+                .WithInnerException<XmlException>()
+                .Where(ex => ex.Data["Saml2Response"] as string == "<foo />");
+        }
+
+        [TestMethod]
+        public void AcsCommand_Run_ResponseIncludedInException()
+        {
+            string payload =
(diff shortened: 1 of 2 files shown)


Kentor.AuthServices.Tests/WebSSO/AcsCommandTests.cs
```diff
@@ -86,7 +86,31 @@ public void AcsCommand_Run_ErrorOnIncorrectXml()
...
+                .Where(ex => ex.Data["Saml2Response"] as string == "<foo />");
+        }
+
+        [TestMethod]
+        public void AcsCommand_Run_ResponseIncludedInException()
+        {
+            string payload =
+                @"<saml2p:Response xmlns:saml2p=""urn:oasis:names:tc:SAML:2.0:protocol""
+                xmlns:saml2=""urn:oasis:names:tc:SAML:2.0:assertion""
+                ID = ""AcsCommand_Run_SuccessfulResult"" Version=""2.0"" />";
```
AndersAbel: Change the ID to be the name of the test, to avoid problems with replay detection.

albinsunnanbo:
Pull request updated

AndersAbel:
Merged in 38bd133

Issue #96, opened 2014-06-25
WebAPI support


henningjensen:
Hi!

I'd like to use this library together with the ASP Web API framework. I'm building a javascript client using Web API controllers as backend. 

Currently the System.Web.Http.AuthorizeAttribute doesn't work like System.Web.Mvc.AuthorizeAttribute. I only get a HTTP 401 - Unauthorized. I need to get 303 together with the authentication url. 

Is it much work to get support for Web API? I might get some time to do this myself if you can point me in the right direction.

PR #228, opened 2015-04-28
Add Web API controller to OWIN sample application

cornem:
Adds a secure web API controller to the OWIN sample application to show authentication works for web API (see issue #96).


AndersAbel:
Thanks a lot @cornem for not only testing, but also adding to the sample application.

PR #230, opened 2015-05-04
Update Configuration.md

DDtMM:
The kentor.authServices element was written mistakenly as an empty element.


AndersAbel:
Thanks for noticing and making a PR. It's already fixed by #236 though (according to the numbering you were first, but I somehow ended up merging the other one) so I'll close this one.

PR #231, opened 2015-05-07
Stubidp: Focus NameId field and select text to make it faster to type a new id

albinsunnanbo:


Kentor.AuthServices.StubIdp/Views/Home/Index.cshtml
```diff
@@ -53,4 +53,14 @@
...
\ No newline at end of file
+}
+
+
+<script>
+    // Focus first relevant field to use
+    if ($("#InResponseTo").val()) {
+        $("#NameId").focus().select();
+    } else {
+        $("#InResponseTo").focus();
```
AndersAbel: For unsolicited responses (where the login is initiated from the Idp) the InResponseTo should be blank, but in that case the ACS URL can't be detected (as there is no AuthNRequest to extract it from).

PR #236, opened 2015-05-12
Fixing bad sample configuration XML

ronaldscott:
Modifying erroneous self-closing element.


AndersAbel:
Thanks a lot for noticing and fixing.

PR #247, opened 2015-06-17
Proposed to fix #246: Read Audience Uris from web.config/system.IdentityModel. 

### LINKED ISSUE #246: Read Audience Uris from web.config/system.IdentityModel.
I'm having an issue where Saml2PSecurityTokenHandler.cs is throwing an exception (which internally is coming from 'base.ValidateConditions'): 

ID1038: The AudienceRestrictionCondition was not valid because the specified Audience is not present in AudienceUris.

Audience: 'localhost:54295'

I've tried various permutations of settings in my web.config under 'system.identityModel > identityConfiguration':

      <audienceUris>
        <add value="http://localhost:54295/" />
      </audienceUris>

I've tried these values:

http://localhost:54295/
http://localhost:54295
localhost:54295
localhost:54295/

No matter what I set, it still tells me 'Audience is not present in AudienceUris'. 
I've also tried 'audienceUris mode="Never"' which doesn't seem to have any effect.

Am I doing somthing wrong?

mip1983:
Update to constructor on Saml2PSecurityTokenHandler to read audience restrictions from system.identitymodel and fall back to a restriction based on entityid.id if none have been applied. Covered by two tests in Saml2PSecurityTokenHandlerTests:

Saml2PSecurityTokenHandler_ShouldHaveDefaultAudienceRestrictionOfEntityId
Saml2PSecurityTokenHandler_ShouldReadAudienceUrisFromIdentityModelConfig

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -9,3 +9,4 @@ Sebastian Allard
 Jozef Raschmann
 Tor-Björn Holmström
 Stephen Patches
+Matthew Paul
--- a/doc/Configuration.md
+++ b/doc/Configuration.md
@@ -87,7 +87,8 @@ Root element of the config section.
 The name that this service provider will use for itself when sending
 messages. The name will end up in the `Issuer` field in outcoing authnRequests.
 
-The `entityId` should typically be the URL where the metadata is presented. E.g.
+The SAML standard requires the `entityId` to be an absolut URI. Typically it should
+be the URL where the metadata is presented. E.g.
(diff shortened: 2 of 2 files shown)


AndersAbel:
Sorry for taking so long before I reviewed this. I fixed some issues when setting the audience validation mode to never (was not properly tested, so I added a test first).

AndersAbel:
Note: This is a breaking change, as configuration is now required to be set to AudienceMode.Never to accept assertions that do not contain an AudienceURI (such as the ones from the stub-idp).

PR #252, opened 2015-06-24
Configurable user lists

albinsunnanbo:
Add support for named IDP:s (with Guid as name) within the StubIdp, each with a configuration file, including a user list.

Add a small list of predefined users on the default IDP.


albinsunnanbo:
This is based on #245, I will rebase this PR if you merge #245 first.

PR #258, opened 2015-07-01
Fix Compilation warning "The variable 'ex' is declared but never used"

albinsunnanbo:
Fix Compilation warning "The variable 'ex' is declared but never used" and corresponding code analysis warning


AndersAbel:
Thanks for noticing. I added that when debugging and fixing the code coverage, but forgot to remove it.

albinsunnanbo:
I can see exactly how it happened.
See https://msdn.microsoft.com/en-us/library/ms164891.aspx and learn the pseudo variable `$exception`.
Enter `$exception` in the watch or quick watch window instead.

Not useful here, but `(System.Data.Entity.Validation.DbEntityValidationException)$exception` is a favorite when debugging entity framework.

PR #259, opened 2015-07-01
Fixes invalid content-length bug in CommandResultExtensions on Auth redirect

JoeBrockhaus:
This was a maddening multi-day bug. 
We are using WebApi & IdentityServer3 with Kentor for SAML2 Auth. 
It worked most times on a W10 VM, but failed on 3 other 8.1 VMs. 
We were never able to figure out why the W10 VM worked, when it did, but after this change it never failed again.  

The root issue is that an HTTP response with Content-Length = 0, but with a non-zero-length body is invalid and clients reject the response or generally exhibit inconsistent behavior. The second issue was helping to obscure the problem. 

1. Content-Length must be set PRIOR to writing to the response stream, because it will send it to the client immediately. 

2. Using a StreamWriter to write to context.Response.Body is unnecessary when context.Response.Write("") exists. In addition to being unnecessary, it adds 3 bytes (in our tests) to the response body. When Content-Length *was not explicitly* set before writing to the response, the response would include *only* those 3 errant bytes (It would still be set to 0). And when a StreamWriter was used & the Content-Length *was set* to the length of the byte array of the desired content, the client still received an invalid response because the difference in header vs body length, by those 3 errant bytes. (That said, in this instance in IE, the browser rendered the redirect content) 

invalid http response


albinsunnanbo:
Thanks for the detailed analysis, we'll have a look at it.

AndersAbel:
Thanks for finding and analyzing this. A question: Are you using the http POST binding when redirecting the user to the idp? I don't think we've used POST in production, so that can explain why we've not seen the problem. There is however another bug #193 that relates to how content is written to Owin responses that we've seen and might be related to this.

As you might have seen we're lagging behind on releases. We're actually just now wrapping up the 0.13.0 release. Unfortunately we won't be able to get this into it, but my ambition is to have more frequent releases again so it shouldn't take too long before this is in a release.

JoeBrockhaus:
Thanks for the quick reply guys. 
For the time being we are just using the forked code in our repo. So while it's not an immediate need, it would be nice to revert back to a NuGet dependency for this.

AndersAbel:
I've taken the change to use `context.Response.Write` into the master branch, but not the rest as I cannot reproduce the issue. I would like to have more information before adding the code that manually sets the Content-Length. Please see discussion in #295, which is a bug reported on this issue.

Kentor.AuthServices/Federation.cs
```diff
@@ -136,13 +136,10 @@ private void RegisterIdentityProviders(List<IdentityProvider> identityProviders)
...
 
-            // Remember what we registered this time, to know what to remove nex time.
-            foreach (var idp in identityProviders)
-            {
-                registeredIdentityProviders = identityProviders.ToDictionary(
-                    i => i.EntityId.Id,
-                    i => i.EntityId);
-            }
+            // Remember what we registered this time, to know what to remove nex time.            
+            registeredIdentityProviders = identityProviders.ToDictionary(
```
JoeBrockhaus: The dictionary only needs to be built once.

Kentor.AuthServices/Federation.cs
```diff
@@ -136,13 +136,10 @@ private void RegisterIdentityProviders(List<IdentityProvider> identityProviders)
...
             }
 
-            // Remember what we registered this time, to know what to remove nex time.
-            foreach (var idp in identityProviders)
-            {
-                registeredIdentityProviders = identityProviders.ToDictionary(
-                    i => i.EntityId.Id,
-                    i => i.EntityId);
-            }
+            // Remember what we registered this time, to know what to remove nex time.
```
JoeBrockhaus: The dictionary only needs to be built once, not *n* times.

Kentor.AuthServices/Federation.cs
```diff
@@ -136,13 +136,10 @@ private void RegisterIdentityProviders(List<IdentityProvider> identityProviders)
                 options.IdentityProviders.Remove(idp);
             }
 
-            // Remember what we registered this time, to know what to remove nex time.
```
JoeBrockhaus: The dictionary only needs to be built once, not *n* times.

JoeBrockhaus:
Removed the explicit length setting.

AndersAbel:
The remaining fix (except whitespace changes and an unrelated change to an exception) has already been done in 6de45845. Closing this.

Issue #201, opened 2015-03-04
Support for local storage of metadata


AndersAbel:
Support local paths for the metadata location for federations and identity providers.

BREAKING: The configuration option is renamed to MetadataLocation.

gerhard-j:
I vote for this feature!

PR #274, opened 2015-07-24
Support for loading metadata from local xml files using file:// URI scheme

CryptonZylog:
Current code loads any metadata xml from a remote URI using WebClient. This will add support to loading the metadata xml file locally in certain SAML integration scenarios. For example, the metadata XML can be stored in the App_Data directory.

This might be a resolution for #201 

--------------------

InvalidCastException happening in LoadFederation() because underlying Load() returns an ExtendedEntityDescriptor, so a wrapper added that just creates a ExtendedEntitiesDescriptor with a single ExtendedEntityDescriptor inside

### DIFF
--- a/Kentor.AuthServices/Configuration/FederationElement.cs
+++ b/Kentor.AuthServices/Configuration/FederationElement.cs
@@ -14,7 +14,7 @@ namespace Kentor.AuthServices.Configuration
     {
         private const string metadataUrl = "metadataUrl";
         /// <summary>
-        /// Url to download metdata for the federation from.
+        /// Url to download metadata for the federation from.
         /// </summary>
         [ConfigurationProperty(metadataUrl, IsRequired = true)]
         public Uri MetadataUrl
--- a/Kentor.AuthServices/Metadata/ExtendedEntitiesDescriptor.cs
+++ b/Kentor.AuthServices/Metadata/ExtendedEntitiesDescriptor.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.IdentityModel.Metadata;
 using System.Linq;
 using System.Text;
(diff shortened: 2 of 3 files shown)


CryptonZylog:
This might be a resolution for #201

AndersAbel:
@CryptonZylog Thanks for the PR. Unfortunately I'm just about to go offline (hiking in the mountains) for a week, so I won't be able to check this until I'm back to civilization.

dariusdamalakas:
@AndersAbel ping. Is this thread still alive?

AndersAbel:
@dariusdamalakas As alive as anything else in AuthServices... I'm currently working full time on other projects, so I do AuthServices in whatever free time I can manage to find.

Regarding this solution I think that the path should be specified as a local path with support for expansion of "~/" instead of a file:// URL. The most common place for storing data will probably bee "~/App_Data"

AndersAbel:
To get merged, these issues must be fixed:

* There are no tests for the new functionality and test coverage has dropped.
* There is one code analysis warning.
* There are two missing XML comments.

I would also prefer to have support for local paths starting with "~" to support "~/App_Data" for storage, but that is not needed.

AndersAbel:
#201 is now fixed, including support for app relative paths. Closing this.

mnaoumov:
@AndersAbel , this issue should be reopen because InvalidCastException is still happening. a3ccc3643bc476e286b61f4a8b9a47d7b5818421 seems to be a valid fix to use

AndersAbel:
@mnaoumov Metadata that contains a single EntityDescriptor should be referenced as an IdentityProvider from configuration. Metadata that contains an EntitiesDescriptor with nested EntityDescriptor nodes should be configured through the Federation mechanism. Are you referencing a file containing just a single EntityDescriptor from a Federation configuration?

mnaoumov:
@AndersAbel Yes, I reference it from Federation configuration because it is a metadata from AD FS. I might be wrong but I thought that if "F" stands for "Federation" I need to register it in Federation configuration. It might be useful to explain that in documentation more clearly for SAML newbies like me :)

AndersAbel:
@mnaoumov ADFS names it as Federation, but it as actually just a single EntityDescriptor in the document. Register it as an IdentityProvider instead. If you have some suggestions for the documentation you are welcome with a PR, otherwise I'll try to remember to add it later when I updated documentation.

PR #275, opened 2015-07-24
Minor fixes to documentation

explunit:


AndersAbel:
Thanks a lot. It's very valuable to have a native English speaker proof-reading things (we're all Swedish speaking and have English as a second language).

PR #280, opened 2015-07-29
line ending issue in string constants

sonicsqwirl:
fixed line endings in expected xml so tests pass


AndersAbel:
Changing line endings in a constant with line breaks is brittle. Please insert explicit \n instead:

    "Some long\n" +
    "string";

That's more reliable as it doesn't risk to change when someone opens it with other line ending settings.

AndersAbel:
Closing due to inactivity.

PR #281, opened 2015-07-30
Fix code analysis issue

explunit:
The unused variable `idp` is detected by VS2015 code analysis but was not detected by VS2013


albinsunnanbo:
Thanks for the PR.
I recognize this somehow, maybe we fixed a similar issue a while ago.
Currently there is vacation season here in Sweden, but we'll merge this as soon as we are back.

Issue #132, opened 2014-10-08
Fails on redirect binding when destination contains query


AndersAbel:
If the destination URI of a redirect binding contains a query string, the binding fails.

First noted by @panmanphil in comment on #119.

bomzj:
When IDp sso url contains query string, Kentor generates url with second question mark. Is there any progress on the issue ?

albinsunnanbo:
I'm not aware of anyone working on this issue, you are more than welcome to provide a pull request if you have the time.

AndersAbel:
Fixed by #296

PR #296, opened 2015-08-22
Fix for #132 (Fails on redirect binding when destination contains query)

Peperud:


AndersAbel:
Code looks good, but a test case is needed before merging. I assume that the old test cases shows that it works when there isn't a query-string, so they should be kept as is. Then a new test case should be added that handles the case when there is a query string.

AndersAbel:
Thanks for that. A small tip for next time: Don't do a PR from master, as there is a risk that things get messed up.

PR #302, opened 2015-09-13
Remember last selected user in the StubIdp

albinsunnanbo:
Save selected user in a cookie, restoring last selected user when returning to the StubIdp.
Keeping a separate cookie for each sub-IDP to make them independent if you use more than one StubIdp.

I wrote this code as a test to see if it's useful or not. I can't really come to a conclusive decision.
On one hand it saves a few clicks/keystrokes when you mostly use your favorite user all the time, but on the other hand it's more complicated to use when you want to change user, especially since I moved focus to the Submit button if restoring the selected user was successful.


albinsunnanbo:
@banansson or @hsahlberg may have an opinion?

Kentor.AuthServices.StubIdp/Scripts/ViewIndex.js
```diff
@@ -98,4 +100,21 @@
             resetUnobtrusive($("form"));
         }
     });
+
+    var cookiedName = 'stubIdp.username';
```
AndersAbel: Why the 'd' in the name of the cookie?

AndersAbel:
I think that this makes sense, so I'm willing to merge it.

Kentor.AuthServices.StubIdp/Scripts/ViewIndex.js
albinsunnanbo: Just making sure you review the code properly.

albinsunnanbo:
Fixed the spelling

banansson:
Late reply, but I believe this would actually save us a few keystrokes more then be disruptive. I am all for.

AndersAbel:
@banansson You're late to the party. It's already merged - and should be live on http://stubidp.kentor.se as it uses continuous delivery.

Issue #304, opened 2015-09-17
Using Okta as IdP within IdentityServer3 v2


dahlsailrunner:
Hi - 
I'm trying to get Okta (a SAML-based corporate SSO provider) configured as an IdP configured in IdentityServer3 using the Kentor.AuthServices (and Owin).

My code is as follows:

    var authServicesOptions = new KentorAuthServicesAuthenticationOptions(false)
    {
                SPOptions = new SPOptions
                {
                    EntityId = new EntityId(my audience value from okta)
                },
                SignInAsAuthenticationType = signInAsType,
                AuthenticationType = "okta",
                Caption = "Okta",
            };

            authServicesOptions.IdentityProviders.Add(new IdentityProvider(
                new EntityId(metadataaddressfromokta)
                authServicesOptions.SPOptions)
                {
                    LoadMetadata = true,
                });

            app.UseKentorAuthServicesAuthentication(authServicesOptions);            

The problem is almost certainly in the way I've got things configured, but I'm not sure where the problem is.  The SPOptions Entityid is the value I have set for the "audience" within my developer instance at Okta.  The Metadata address in the IdentityProvider constructor is the metadata address which I got from the okta config page.

The detailed error is shown below.

Any insight is greatly appreciated.
Thanks - 
Erik

Unexpected entity id "http://www.okta.com/exk4yxtgy7ZzSDp8e0h7" found when loading metadata for "https://MyOktaInstance.oktapreview.com/app/exk4yxtgy7ZzSDp8e0h7/sso/saml/metadata". 
  Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

 Exception Details: System.Configuration.ConfigurationErrorsException: Unexpected entity id "http://www.okta.com/exk4yxtgy7ZzSDp8e0h7" found when loading metadata for "https://MyOktaInstance.oktapreview.com/app/exk4yxtgy7ZzSDp8e0h7/sso/saml/metadata".

Source Error: 

Line 82:             };
Line 83: 
Line 84:             authServicesOptions.IdentityProviders.Add(new IdentityProvider(
Line 85:                 new EntityId("https://MyOktaInstance.oktapreview.com/app/exk4yxtgy7ZzSDp8e0h7/sso/saml/metadata"),
Line 86:                 authServicesOptions.SPOptions)

AndersAbel:
[...]

PR #308, opened 2015-09-28
Oktadoc

dahlsailrunner:
Based on Albin Sunnanbo's request from issue #304, I have added some documentation regarding what I did to get Okta configured as an identity provider within my IdentityServer3 project.  I added a new doc to the doc folder, an image to same with a screenshot of config options, and updated the main readme to ref the new doc.


albinsunnanbo:
Hi, thanks for the effort!
It looks good and readable to me.

Regarding "Audience URI (B)", this probably translates to the AudienceRestriction in SAML2. 
See http://security.stackexchange.com/q/13424/43540 and http://stackoverflow.com/a/5201647/401728
If I recall correctly the EntityID should be the metadata url, in this case “https://id.local/identity/AuthServices”.

dahlsailrunner:
Do you want me to update the doc and do a new pull request, or will you just edit as the merge happens?  I'm happy to edit and resubmit if you like.

albinsunnanbo:
You can just push updates to your branch and they will be included in the pull request.
The pull request is a "live" pointer to the head of your branch.
Just send a message here when you´re done.

AndersAbel:
Thanks for the extensive documentation. This will be valuable both for those using IdentityServer3 + Okta and also for general setup of AuthServices with IdentityServer3.

You're (nearly) right in that the EntityId just have to be a prearranged value between the the Okta application and the IdentityServer. It can't be any string though, it has to be an absolute URI. The convention is to use the Url of the metadata of the SP, which in the example would be https://id.local/identity/AuthServices/.

As @albinsunnanbo noted, you can push another commit to your branch which will update the PR. However there is no built in notification in GitHub when you do so, so please add a comment here to make sure we see the change.

dahlsailrunner:
I've updated the doc regarding the audience URI (thanks for the input and clarification on this).  Let me know if you see any other edits that should be made.

AndersAbel:
Thanks for excellent documentation.

PR #316, opened 2015-10-23
for some reason the metadata return more than one signing key (ADFS), this …

xxjthxx:
…caused an exception


albinsunnanbo:
Hi
Thanks for the effort to fix, however it does not solve the real issue.
When there are more than one certificate present in the metadata this indicates a pending certificate switch and the IDP may at any moment stop using the first and start using the second. Any certificate in the list could thus be the one the IDP is using. Just checking the first is not enough.

This is a known issue #105

AndersAbel:
I'm actually working on this, see the [105MultipleIdpKeys](https://github.com/KentorIT/authservices/tree/105MultiIdpKeys) branch.

PR #322, opened 2015-10-28
Restructuring of how Pending Authentication data is stored

akiander:
Exposes a new Interface that allows consumers to easily swap out their own implementation to store pending authentication data in an alternative storage mechanism, such as a database.

Consumer just has to hook up their own container via options... for example: 

            var options = new KentorAuthServicesAuthenticationOptions(false)
            {
                SPOptions = new SPOptions
                {
                    EntityId = new EntityId("MyEntityId")
                },

                AuthenticationType = "saml2p",
                Caption = "saml2p",
                SignInAsAuthenticationType = signInAsType
            };

            options.IdentityProviders.Add(new IdentityProvider(
              new EntityId("IdpEntityId"),
              options.SPOptions)
            {
                MetadataUrl = new Uri("Metadata_Uri"),
                LoadMetadata = true, 
                PendingAuthStorageContainer = new MyPendingAuthDbStorage()  //hook up your own IPendingAuthStorageContainer here
            });

            app.UseKentorAuthServicesAuthentication(options);


martinjt:
I think that you should be a little more explicit about the InMemory option,  maybe call it "Static".

Also not sure on the term "container", maybe just IPendingAuthnRequests.  That's just me.

Other than that, it's perfect for what I need.  Great work!

akiander:
I used "InMemory" just as a way to emphasize that it won't work in a multi-server environment. That's the key problem I have, which is solved with this check-in. The nice thing about the strategy employed here is that it allows clients to provide their own technique for storing this information in whatever storage mechanism that makes sense for them. 

I did try to employ the cookie-based storage technique recommended in [Issue 98](https://github.com/KentorIT/authservices/issues/98) but it ran into conflicts with the core Identity Server 3 code that I am plugging into, so I abandoned that. 

Now I just store the state in SQL Server using Entity Framework by just plugging in a class that inherits from new interface in this pull request.

martinjt:
I completely understand why it's needed, and I'm behind the idea you're done (I suggested it but not had chance to work on it).

I'm not bothered about the cookie solution personally.  I'll be implementing a Redis version for our environment.

I'm just thinking that the naming needs to be write... probably best for @AndersAbel to weigh in.

akiander:
Agreed - I would be happy to rename any of these types. I'll wait for @AndersAbel to weigh in.

albinsunnanbo:
I like the pluggable approach. The implementation looks reasonable small and uncomplicated.
Some documentation in doc/Configuration*.md would be good.

Is the SQL adapter you have build general enough to be submitted as an official plugin?
I found the approach used by Hangfire reasonable: using a separate schema for the tables, and trying to create them automatically if they don't exist. They provide a link to the SQL script for creating the tables in the documentation if your application does not have permission to create tables.

AndersAbel:
Please don't submit a pull request from your own master branch. That will always cause pain in the end. See https://coding.abel.nu/2015/03/fixing-a-pull-request-from-master/ for how to fix it.

Kentor.AuthServices/IPendingAuthStorageContainer.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Storage container for authentication state that must be present when an authentication response is received back from the Identity Provider.
+    /// </summary>
+    public interface IPendingAuthStorageContainer
+    {
+        /// <summary>
+        /// Add state that must be available after external authentication response comes back from IdP. 
+        /// </summary>
+        /// <param name="id"></param>
```
AndersAbel: Document the params.

Kentor.AuthServices/IdentityProvider.cs
```diff
@@ -130,6 +130,31 @@ public Saml2BindingType Binding
...
         }
 
+        private IPendingAuthStorageContainer pendingAuthStorageContainer;
+
+        /// <summary>
+        /// How pending authentication metadata is stored before sending Authentication Requests 
+        /// to the identity provider and retrieved again after a response is returned. Defaults
+        /// to an in-memory container. 
+        /// </summary>
+        public IPendingAuthStorageContainer PendingAuthStorageContainer
```
AndersAbel: I don't think that this belongs on the IdentityProvider level. It should probably be part of ISPOptions, or maybe IOptions.

Kentor.AuthServices/Internal/PendingAuthInMemoryContainer.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Kentor.AuthServices.Internal
+{
+    /// <summary>
+    /// In memory container for storing in-progress Authentication requests. 
+    /// </summary>
+    internal static class PendingAuthnInMemoryContainer
```
AndersAbel: Remove this class. It doesn't make sense to have a class that just wraps a static dictionary. Better to go all the way and have an instance as part of IOptions or ISPOptions and get rid of `static` completely. That will make the isolation better if multiple instances of the Owin middleware are used in the same process too.

AndersAbel:
Overall this looks like a good approach, but there are some code standard things that need to be fixed:

* Some of the tests (e.g. PendingAuthnRequests_AddRemove) should be renamed and moved to a new file, as they are no longer testing the PendingAuthnRequests class but instead testing the new PendingAuthInMemoryStorage.
* There is one failing test.
* Code coverage is not complete. If you don't have VS Enterprise, let me know and I can tell you exactly what lines are missing coverage.

And by the way, I think that the class should be named `PendingAuthnInMemoryStorage`. It's fine with "InMemory" to me, but I think that the "n" should be kept in the name.

There are some things that need to be improved before merging. You will probably want to do a rebase on master too to keep up to date. Unfortunately that is not possible with a pull request from your own master. You will have to create a branch and make a new PR from that. I will close this pull request now and am looking forward to the new one.

For some git tips: please see https://coding.abel.nu/2015/03/fixing-a-pull-request-from-master/ and https://coding.abel.nu/2015/10/become-a-git-wizard-with-7-simple-tricks/

akiander:
I appreciate the feedback. I'll have a new pull request ready as soon as
possible.

PR #323, opened 2015-10-29
Updated Okta document for IdP-initiated login notes

dahlsailrunner:
Hi guys -- I updated the Okta documentation to include a new step 5 to setup IdP-initiated logins.  That's the only change I made -- just the one Markdown file.  Let me know if this looks ok or if there is another way you'd like me to do this.

### DIFF
--- a/doc/IdentityServer3Okta.md
+++ b/doc/IdentityServer3Okta.md
@@ -103,3 +103,65 @@ public class Startup
 
 ## Step 4 - Try it out!  
 Logging in with Okta through your identity server should work at this point.  Cheers!
+
+## Step 5 - Set up IdP-Initiated support 
+This is an optional but very nice and (I think) important step that will enable Okta users
+to login to your site by clicking on the "app" icon on their Okta Dashboard.  The process
+involves setting up the Kentor options to "allow unsolicited requests", and setting up a
+redirect page on your target site (not identity server), and specifying that in the Kentor
+settings.  Explanations follow.
+
+The first step is to configure the AuthServices/Acs endpoint to allow for unsolicited requests.
+To do this, modify the code snippet where you are adding IdentityProviders to include the 
+AllowUnsolictedAuthnResponse = true line shown below.
+
+```csharp
+			authServicesOptions.IdentityProviders.Add(new IdentityProvider(
+                new EntityId(oktaEntityId), authServicesOptions.SPOptions)
+            {
+                LoadMetadata = true,
+                MetadataUrl = new Uri(oktaMetadataUrl),
+                AllowUnsolicitedAuthnResponse = true
+            });
+```
+
+Now you need to set up your target website (not IdentityServer3) to have a page that will
+simply turn around and redirect to your identity server with an authorize request.
+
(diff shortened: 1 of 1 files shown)


doc/IdentityServer3Okta.md
```diff
@@ -102,4 +102,66 @@ public class Startup
...
 
 ## Step 4 - Try it out!  
-Logging in with Okta through your identity server should work at this point.  Cheers!
\ No newline at end of file
+Logging in with Okta through your identity server should work at this point.  Cheers!
+
+## Step 5 - Set up IdP-Initiated support 
+This is an optional but very nice and (I think) important step that will enable Okta users
+to login to your site by clicking on the "app" icon on their Okta Dashboard.  The process
+involves setting up the Kentor options to "allow unsolicited requests", and setting up a
```
AndersAbel: Should be "allow unsolicited response".

doc/IdentityServer3Okta.md
```diff
@@ -102,4 +102,66 @@ public class Startup
...
+redirect page on your target site (not identity server), and specifying that in the Kentor
+settings.  Explanations follow.
+
+The first step is to configure the AuthServices/Acs endpoint to allow for unsolicited requests.
```
AndersAbel: reponse.

AndersAbel:
Looks good, but I have one concern (except for the small typos) and that's the risk of a CSRF attack. What if a user is redirected to the ReturnUrl by a malicious site?

I think that the risk should be mentioned. It's hard to make a workaround without doing some major rewriting of AuthServices, so for now it can be there. In any case, the important CSRF protections is on sensitive operations in the client application.

Also, please don't submit pull requests from your master branch - make a separate branch. For this one it can be handled - but update your master branch from the upstream master until the PR is merged.

dahlsailrunner:
I will refork and resubmit from a branch off of my forks' master.  I need the practice anyway.  :)
Regarding CSRF -- if someone went to my "IdP_InitiatedRedirect.aspx" page -- even with the idp=okta query string, I *think* we would be protected because that will be making an "authorize" request through both Identity server and then my custom user service.  If Okta (the IdP does not recognize a logged-in user, they will get stuck at a login screen.  If the user is not valid in my application (cannot look up roles, etc), then they won't get past my custom user service.  So I think we are safe.  I can add some notes to this effect in the doc if you like.
Let me know your thoughts and you can "reject" this PR and I will submit a new one.  Sorry for the trouble.

AndersAbel:
You don't have to refork. Create a new branch in your repo from your current master (git checkout master; git checkout -b IdpInitiatedOkta). Then add KentorIT/AuthServices as another remote (called "upstream") to your local repo and reset your master to upstream/master (git remote add upstream https://github.com/KentorIT/AuthServices; git checkout master; git reset --hard upstream/master; git branch -u upstream/master) The final command will set your master to track the master in the upstream (i.e. master) repo. See also https://coding.abel.nu/2015/10/become-a-git-wizard-with-7-simple-tricks/ and https://coding.abel.nu/2015/03/fixing-a-pull-request-from-master/

I'm closing this PR as it is tied to your master branch. When you create a new branch, you will have to create a new PR.

PR #324, opened 2015-10-29
added some notes regarding setting up IdP-initiated login from Okta.

dahlsailrunner:
I *think* this will be a better branch to pull from.  I've fixed the typos and did *not* add anything regarding CSRF.  Let me know if you think we should add anything.  And thanks for the GitHub tips -- I'll review them for sure.


AndersAbel:
I think we can leave the CSRF issue for now. It's actually not anything worse than what happens if I know that you are logged in to a specific application. The application must protect itself from that scenario, meaning any CSRF issues in this setup won't matter in the end. I'll merge this.

And yes, that's a better branch to do make the pull request from.

PR #326, opened 2015-11-05
Clients can change how Pending Authentication data is stored

akiander:
This is the second attempt at implementing a mechanism for clients to provide their own mechanism for storing pending authentication data. Primary reason for doing this is because the default mechanism stores data in memory which won't work in a multi-server environment. 

Improvements made here include many (most?) of those requested from the [original pull request #322](https://github.com/KentorIT/authservices/pull/322). 

I do not have the code coverage tools to analyze which lines of code may not have coverage, so if you let me know what those are, then I can repair it and resubmit. 

I did my best to build on a branch that is properly rebased, but let me know if anything is wrong and I'll try again to correct it.


jhoerr:
Hi, @akiander -- we'd really like to use `authservices` but this one issue is holding us back. Do you have a sense for when it might be reviewed?

jhoerr:
Hi, @AndersAbel -- we'd really like to use `authservices` but this one issue is holding us back. Do you have a sense for when it might be reviewed?

Peperud:
@AndersAbel 
Another vote for a load balancer friendly Kentor.

crixo:
@AndersAbel 
Adding my vote too.

AndersAbel:
Ok, I'm feeling the pressure :-)

I'm currently working on single logout which has a hard deadline from a paying customer (unfortunately I have to pay my bills too, so money has more weight than votes).

Regarding the solution: I've made some initial work to move the entire pending state out of memory and instead preserve it in a cookie. That will both solve the problem and give proper CSRF protection. What is your opinion? Do we need the flexibility to chose an implementation as suggested in this PR, or would it be enough to go with a cookie based solution?

crixo:
@AndersAbel first of all thanks a lot for all your effort: the library is simply great!
I don't think anyone what to be pushing and we all understand the priority so no one what to mess the backlog :-)
I guess we were simply pointing out the risk to go live with the current version in a load balancer scenario that I assume is a very common scenario.
I did not have dig to much in the code.protocol itself to understand the data that stored within the storage. If is not predictable and the response will be validate itself according to cert and trus, I guess the correlation id may stored within a cookie as first step. For sure make the storage puggable is the way to go.

akiander:
A cookie-based solution is fine with me.

jhoerr:
A cookie-based solution is fine with us as well. Would we continue to mark that cookie as secure with the `cookieHandler` web.config element?

```
<system.identityModel.services>
  <federationConfiguration>
    <cookieHandler requireSsl="true" />
  </federationConfiguration>
</system.identityModel.services>
```

AndersAbel:
@jhoerr No, The config section you are referring to is the one used by the SessionAuthenticationModule (used by MVC and HttpModule packages). AuthServices will automatically set "secure" based on the protocol of the PublicOrigin URL.

AndersAbel:
In the single logout feature I've done some work on cookie handling that can be reused for this. During that work I also discovered that a server side store without a cookie **opens up for a CSRF attack**. To properly protect against CSRF scenarios (which is need e.g. when linking external authentication options with the ASP.NET Identity template) there must be a cookie. That means that this PR will be closed and not merged, in favour for a cookie based solution.

jhoerr:
Thanks for the update @AndersAbel. Any ETA on a cookie-based solution?

PR #327, opened 2015-11-09
Lazy loading of metadata

xxjthxx:
- this loads metadata only if the provider is used. Might be useful if some providers are unreachable and waiting for them to timeout causes unnecessary blocking on the start

### DIFF
--- a/Kentor.AuthServices.Tests/App.config
+++ b/Kentor.AuthServices.Tests/App.config
@@ -32,6 +32,7 @@
         <signingCertificate fileName="Kentor.AuthServices.Tests.pfx" />
       </add>
       <add entityId="http://localhost:13428/idpMetadata" allowUnsolicitedAuthnResponse="true" loadMetadata="true" />
+      <add entityId="http://localhost:13428/idpLazyMetadata" allowUnsolicitedAuthnResponse="true" loadMetadata="false" lazyLoadMetadata="true" />
     </identityProviders>
     <federations>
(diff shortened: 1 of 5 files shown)


xxjthxx:
What is the status of this pull request?

AndersAbel:
Status is that I finally have some time to work on AuthServices, so I'm reviewing PRs today.

I'm trying to understand what scenario you are trying to handle here. Is the problem that the metadata is loaded synchronously for all idps on startup? With one idp being stuck things get slow?

In that case I think it would be better to load the metadata asynchronously instead for all Idps. Is it really needed to offer this as a configurable value? Having a smaller config interface makes it easier for the users.

Another thing that might be relevant is that I plan to make an on-disk-cache of metadata to improve startup performance. When having that, this might even be redundant.

What do you think? If you have any real life scenario that prompted you to make this update I'd love to hear what happened. It's always better to know concrete scenarios when planning features.

xxjthxx:
Hello!

> I'm trying to understand what scenario you are trying to handle here. Is the problem that the metadata is loaded synchronously for all idps on startup? With one idp being stuck things get slow?

That's exactly the problem I was having. My change seemed to me the most straight-forward answer, but after reading what you wrote and  rethinking it I think I could do without it, so as you said it would be better not do include additional switch.

All in all, we can drop this :-). 

Having said that what I dislike here is the way loading metadata is enforced when the properties are set in the IdentityProvider class. In my opinion this is a bad design. Setter should not explicitly trigger a network call. The easiest way to deal with this would be to abstract an interface to load meatadata and have only a dependency to the interace, not the actual implementation (which might be loading metadata from the network).

PR #331, opened 2015-11-16
added configuration property to allow override the Acs base url

simon-hofer:
Hi,
We have the problem that one of our customers serves the webapplication using the authservices behind a reverse proxy on diffrent url than in web. This causes a wrong saml2p:AuthnRequest with wrong AssertionConsumerServiceURL.
In this pull request I added an optional configuration property to authservices section to override the default behaviour and set the base url of the application to use in the AuthnRequest.

Thanks a lot for your very helpful framework!

Kind Regards,
Simon


AndersAbel:
This is something that needs to be added, but the solution need to be extended:

* The configuration have to follow the conventions for the rest of the system, where everything is done through ISPOptions which allows both configuration through web.config and from code.
* The application of the configuration have to be moved out of the HttpModule and into somewhere in Core, to make it available for all three API modules. If it's not possible to move to Core, it has to be implemented in the MVC module and Owin too.

explunit:
Minor point regarding naming of the config option: `assertionConsumerServiceUrl` is more specific than what you're actually doing here. IdentityServer3 uses `PublicOrigin` as the name for a very similar option. https://identityserver.github.io/Documentation/docs/configuration/identityServerOptions.html

Kentor.AuthServices.Tests/WebSSO/SignInCommandTests.cs
```diff
@@ -134,5 +137,27 @@ public void SignInCommand_Run_ReturnsRedirectToDiscoveryService()
...
+
+        [TestMethod]
+        public void SignInCommand_Run_MapsAssertionConsumerServiceURL()
+        {
+            KentorAuthServicesSection.Current.AllowChange(true);
+            KentorAuthServicesSection.Current.AssertionConsumerServiceUrl = new Uri("https://testurlForAcs.com/");
+
+            try
+            {
+                Assert.IsNotNull(KentorAuthServicesSection.Current.AssertionConsumerServiceUrl);
```
AndersAbel: Please use FluentAssertion instead of Assert. to follow the convention of the rest of the tests.

simon-hofer:
I've tried to update my branch with the suggested changes. PublicOrigin option is added in ISPOptions interface and renamed in the configsection.
I've tried to test for all the modules but it needs some further reviewing.

AndersAbel:
Thanks for the update. The test coverage is perfect, so that's nothing to worry about.

When looking at the code it feels like the change got more wide spread than I imagined would be needed. It's probably due to some complexity that I didn't foresee, but I'd like to have your opinions if another implementation would be possible. All the production code uses the `AuthServicesUrls(HttpRequestData, ISPOptions)` ctor. Would it be possible to move the logic of handling the PublicOrigin to inside that ctor instead of doing it on the HttpRequestData?

simon-hofer:
Yea, I agree and moved the public origin logic to AuthServicesUrls.

Kentor.AuthServices.HttpModule/HttpRequestBaseExtensions.cs
```diff
@@ -5,6 +5,7 @@
 using System.Text;
 using System.Threading.Tasks;
 using System.Web;
+using Kentor.AuthServices.Configuration;
```
explunit: Minor: should be no change to this file now that you've taken the different approach.

Kentor.AuthServices.Owin/OwinContextExtensions.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using Kentor.AuthServices.Configuration;
```
explunit: Same comment for this file

Kentor.AuthServices.Tests/HttpModule/HttpRequestBaseExtensionsTests.cs
```diff
@@ -3,6 +3,7 @@
 using System.Web;
 using Kentor.AuthServices.HttpModule;
 using FluentAssertions;
+using Kentor.AuthServices.Configuration;
```
explunit: Same comment here -- should be no change

Kentor.AuthServices/WebSSO/HttpRequestData.cs
```diff
@@ -8,6 +8,7 @@
 using System.Text;
 using System.Threading.Tasks;
 using System.Web;
+using Kentor.AuthServices.Configuration;
```
explunit: Remove

doc/Configuration.md
```diff
@@ -123,7 +124,7 @@ Defaults to `/AuthServices` if not specified. This can usually be left as the
 default, but if several instances of AuthServices are loaded into the
 same process they must each get a separate base path.
 
-###`authenticateRequestSigningBehavior` Attribute
+####`authenticateRequestSigningBehavior` Attribute
```
explunit: Is this a fix or an accidental change?

simon-hofer: > Is this a fix or an accidental change?

This is a fix. IMHO it was one level to high.

AndersAbel: @simon-hofer you're right. It should be level 4. Thanks for noticing.

AndersAbel:
@simon-hofer Thanks for your contribution and for your patience with our reviews. Thanks also to @explunit for the review work.

PR #332, opened 2015-11-18
Throwing null ref exception if options is null on get claim

dariusdamalakas:
It would make more sense to not check against full exception text, but I am not that good at Fluent assertions. This is what would probably make more sense in unit test: 
````
a.ShouldThrow<ArgumentNullException>()
                .Where(x => x.Message.Should().Contain("options"));
`````

### DIFF
--- a/Kentor.AuthServices/SAML2P/Saml2Response.cs
+++ b/Kentor.AuthServices/SAML2P/Saml2Response.cs
@@ -473,6 +473,10 @@ namespace Kentor.AuthServices.Saml2P
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
         public IEnumerable<ClaimsIdentity> GetClaims(IOptions options)
         {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
             if (createClaimsException != null)
             {
                 throw createClaimsException;
--- a/Kentor.AuthServices.Tests/Saml2P/Saml2ResponseTests.cs
+++ b/Kentor.AuthServices.Tests/Saml2P/Saml2ResponseTests.cs
@@ -1506,6 +1506,38 @@ namespace Kentor.AuthServices.Tests.Saml2P
         }
 
         [TestMethod]
+        public void Saml2Response_GetClaims_ThrowsOnEmptyOptions()
+        {
+            var response =
+            @"<?xml version=""1.0"" encoding=""UTF-8""?>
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by dariusdamalakas** [comment]
Let me know if this looks ok.

**Comment by albinsunnanbo** [comment]
You could use `.And.ParamName.Should().Be("options")`
Look at https://github.com/KentorIT/authservices/blob/4ffe5c54f8b2100a37a6e4cf7269f65a40b0798e/Kentor.AuthServices.Tests/ClaimsIdentityExtensionsTests.cs#L21 for an example.

File: Kentor.AuthServices/SAML2P/Saml2Response.cs [hunk 1]
```diff
@@ -473,6 +473,8 @@ private static void CheckSignature(XmlElement signedRootElement, IEnumerable<Asy
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
         public IEnumerable<ClaimsIdentity> GetClaims(IOptions options)
         {
+            if (options == null)
+                throw new ArgumentNullException("options");
```
albinsunnanbo: https://github.com/KentorIT/authservices/blob/master/CONTRIBUTING.md requires `{}` even for single line `if`.

dariusdamalakas: Thanks, will update this PR now.

dariusdamalakas:
@albinsunnanbo , PR updated.

Kentor.AuthServices/SAML2P/Saml2Response.cs
```diff
@@ -473,6 +473,10 @@ private static void CheckSignature(XmlElement signedRootElement, IEnumerable<Asy
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
         public IEnumerable<ClaimsIdentity> GetClaims(IOptions options)
         {
+            if (options == null)
+            {
+                throw new ArgumentNullException("options");
```
albinsunnanbo: I also recall that we recently switched to use the c#6 operator `nameof` elsewhere in the code.
`throw new ArgumentNullException(nameof(options));`

@AndersAbel this might be useful to mention in the contributors guideline?

dariusdamalakas: @albinsunnanbo , PR updated.

dariusdamalakas:
@albinsunnanbo PR updated to use `nameof` operator.

dariusdamalakas:
@albinsunnanbo , let me know if anything else needs changing

PR #342, opened 2015-12-05
Remove invalid XML declaration from HTML-page in Saml2PostBinding

albinsunnanbo:
I suppose the Saml2PostBinding posting page should not begin with an XML-declaration.


albinsunnanbo:
Seems the `<?xml ?>` declaration is actually recommended for XHTML documents.
You learn something new every day.

PR #350, opened 2015-12-19
StubIDP: Autocomplete in user list

albinsunnanbo:
Implement autocomplete in the User list in the StubIDP.
When the user list passes about 20 users it was complicated to use and find the desired user.
This pull request implements autocomplete in the user list by using the select2 component.
This scales rather well on sample lists of 5000+ users. Both from a performance and a usability perspective.


AndersAbel:
@albinsunnanbo Would you mind rebasing this?

albinsunnanbo:
Rebased, let me know if you want a large dataset for testing

PR #353, opened 2015-12-30
Adding secondary level status code for saml response.

xxjthxx:
As defined in 3.2.2.2 or the Assertions and Protocols for the OASIS Security Assertion Markup Language

- this adds additional string field for the secondary status message, the field is intentionally a string and not enum because specification says this status can change in future releases
- additional exception has been added to get the statuscode and secondary status code if the STS response is other than Success. It may help to handle those situations outside the module.


xxjthxx:
What is the status of this pull request?

Kentor.AuthServices.Tests/Saml2P/Saml2ResponseTests.cs
```diff
@@ -1567,8 +1570,38 @@ public void Saml2Response_GetClaims_ThrowsOnStatusFailure()
...
-            a.ShouldThrow<InvalidOperationException>()
-                .WithMessage("The Saml2Response must have status success to extract claims. Status: Requester.");
+            a.ShouldThrow<InvalidSamlOperationException>()
+                .WithMessage("The Saml2Response must have status success to extract claims. Status: Requester.")
+                .Where(x => x.Status == Saml2StatusCode.Requester);
+
+        }
+
+        [TestMethod]
+        public void Saml2Response_GetClaims_ThrowsOnStatusFailure2()
```
AndersAbel: Please use a more descriptive name. How about Saml2Response_GetClaims_ThrowsOnStatusFailure_IncludingSecondLevelMessage?

Kentor.AuthServices/InvalidSAMLOperationException.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Kentor.AuthServices
+{
+    /// <summary>
+    /// Extended exception containing information about the status and status message SAML response.  
+    /// </summary>
+    [Serializable]
+    public class InvalidSamlOperationException: InvalidOperationException
```
AndersAbel: The exception types used in AuthServices is somewhat non-well-thought-through. This PR made me go back and see what I've done before and it's not very clear. But there is a bas class AuthServicesException that I think that this exception should derive from. What do you think?

AndersAbel:
Please see my comments. Would be great if you could rebase the PR on master as well. I've finally got a build server running so that would give some checks on build, tests and code coverage.

xxjthxx:
Thanks for the feedback, I'll make changes you suggested.

The reason I didn't inherit from AuthServicesException in the first place was that IMO it would be a breaking change (different type of exception thrown), which I feared to introduce as pull request :-). But you're right, it would be best 

What I further propose:
- Move 5 exception classes to separate folder (called asy Exceptions), there's no reason the are in the root directory
- rename InvalidSamlOperationException to UnsuccessfulSamlOperationException. Invalid may make you think that STS is not behaving properly, although it does, it may have responded with login denial.

AndersAbel:
I think both of your proposals make sense, so go ahead with them.

Regarding breaking changes: Now is the time for them! I want to get a 1.0.0 release out soon, so anything breaking should be fixed now, before that. After 1.0.0 it will be much worse with breaking changes.

xxjthxx:
Pull request has been updated.

AndersAbel:
Thanks a lot for the added functionality and for taking the extra time to clean up the exceptions structure.

Issue #355, opened 2016-01-04
Support for service provider certificate rollover


explunit:
I started a branch to support key rollover on the service provider certificate and wanted to validate several assumptions before I go too far with one approach:

Current situation: today we have one `ServiceCertificate` property that's used for decryption of inbound assertions if needed. It can be [read from config](https://github.com/KentorIT/authservices/blob/master/doc/Configuration.md#servicecertificate-element), e.g.

```
<serviceCertificate fileName="Kentor.AuthServices.Tests2.pfx" />
<identityProviders>
...
```

Two basic questions:
 1. Once #95 is implemented, do we want to support using a different certificate for signing requests vs. the one used for decryption?
 1. Do we want to configure the certificates as one list with different properties to indicate their purpose and whether they are the active cert? something like
```
<serviceCertificates>
  <serviceCertificate fileName="cert1.pfx" keyType="Encryption" active="true"  /> <!--published in metadata -->
  <serviceCertificate fileName="cert2.pfx" keyType="Signing" active="true" /> <!--published in metadata -->
  <serviceCertificate fileName="old-cert1.pfx" keyType="Encryption" active="false"  /> <!-- still accepted for use until expired, but no longer published -->
</serviceCertificates>
```

I'm assuming the answer is "Yes" to both but I thought it was worth asking. I'm sure the above example is not exact, but should give an idea of what I was thinking.

And the default would be active=true and keyType is unspecified, so the simple case when used for both signing & encryption and no rollover in progress:
```
<serviceCertificates>
  <serviceCertificate fileName="cert1.pfx"  />
</serviceCertificates>
```

This would be a breaking change vs. 0.14 but I assume this is OK until 1.0.

For comparison:
https://wiki.shibboleth.net/confluence/display/SHIB2/NativeSPMultipleCredentials

AndersAbel:
Thanks for looking into this.

1. Yes, I think we should support it, as well as support a scenario where one certificate is used for both purposes.
2. Yes, that's a good idea.

I like the basic setup of your example config. A few minor thoughts:

* The parent element should be `<serviceCertificates>` as you've done, but the child elements can be named `<add>`. That would follow how federations and identity providers are configured.
* The `<add>` element could subclass the existing `CertificateElement` type that is used for idp certificates. The derived class just needs to add the active and use/keyType flag. That would reuse support for both loading cert store and from file (including web root relative file names).
[...]

PR #358, opened 2016-01-08
Possible refactoring of configuration loading

explunit:
For discussion....

In my opinion, having `KentorAuthServicesSection` implement `ISPOptions` mixes two different concerns in one file (e.g. see the removed lines in this PR that are strictly for `ISPOptions` but don't relate to reading from config file) and make it difficult to have an `ISPOption` property of a different type than we need for reading from config files. Note that `KentorAuthServicesSection` does *not* implement `IOptions`. This lets it to have an `IdentityProviders` property that is of a different type. I'm looking to have that same flexibility e.g. for the `ServiceCertificates` property for #355.

However, I struggled with where to put the "adapter" code that converts from the xml-config-reading objects into the in-memory config objects. For the moment I decided to follow the existing pattern where `IdentityProvider` has constructor that accepts an `IdentityProviderElement`, but I'm open to better ideas on this.

This PR also makes the entire object returned from `FromConfiguration` a singleton rather than just the SPOptions portion of it.

### DIFF
--- a/Kentor.AuthServices/Configuration/Options.cs
+++ b/Kentor.AuthServices/Configuration/Options.cs
@@ -24,13 +24,21 @@ namespace Kentor.AuthServices.Configuration
         {
             get
             {
-                var options = new Options(KentorAuthServicesSection.Current);
-                KentorAuthServicesSection.Current.IdentityProviders.RegisterIdentityProviders(options);
-                KentorAuthServicesSection.Current.Federations.RegisterFederations(options);
-                options.SPOptions.ServiceCertificate = KentorAuthServicesSection.Current.ServiceCertificateConfiguration.LoadCertificate();
-
-                return options;
+                return optionsFromConfiguration.Value;
             }
+        }
+
+        private static readonly Lazy<Options> optionsFromConfiguration 
+            = new Lazy<Options>(() => LoadOptionsFromConfiguration(), false);
+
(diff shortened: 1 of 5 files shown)


Kentor.AuthServices/Configuration/Options.cs
```diff
@@ -24,15 +24,23 @@ public static Options FromConfiguration
...
-                KentorAuthServicesSection.Current.Federations.RegisterFederations(options);
-                options.SPOptions.ServiceCertificate = KentorAuthServicesSection.Current.ServiceCertificateConfiguration.LoadCertificate();
-
-                return options;
+                return optionsFromConfiguration.Value;
             }
         }
 
+        private static readonly Lazy<Options> optionsFromConfiguration 
+            = new Lazy<Options>(() => LoadOptionsFromConfiguration(), false);
```
AndersAbel: I think it should be threadsafe (passing true instead). The config might be accessed by several threads on a busy server.

AndersAbel:
Nice and clean, there's not much to discuss here. Please give me your view on the thread safety flag and then I'm ready to merge.

Kentor.AuthServices/Configuration/Options.cs
explunit: Fixed

PR #361, opened 2016-01-10
Code coverage with codecov

AndersAbel:
Test PR to see how/if it works.


AndersAbel:
Won't merge this, got coveralls working and preferred that, as I can use the Visual Studio coverage tool with that. OpenCover did get somewhat different coverage results than VS which would be confusing.

Issue #95, opened 2014-06-24
Support SP AuthnRequest signing


simongustaf:
Hi, this lib plugged in nicely in my app without any hassle (this is my first atempt with SAML and SSO). I just have one question, are you planning to add signing of the AuthnRequest (with the SP:s certificate)?

Great work guys!
/Simon

cornem:
I'm looking for the same functionality, as we're trying to connect to an IdP that requires a signed AuthnRequest.

I'm going to give this a try myself, if you have not planned this issue yet.

AndersAbel:
Nothing is planned from me on this currently.

If you make a try, please have a look at the comments in #123 and the work in #78 which probably contains some work on adding SP certificates to the configuration.

AndersAbel:
Waiting for #355 to be completed first.

PR #379, opened 2016-01-11
Service provider certificate enhancements

explunit:
Implements #355 -- see details there.

* Support multiple service provider certificates:
  * key rollover
  * separate certificates for signing requests vs decryption of inbound assertions
* Support signing certificates e.g. for #95 

Note: **Breaking change** for anyone who used the `serviceCertificate` property in earlier versions.


Kentor.AuthServices.Tests/Configuration/SPOptionsTests.cs
```diff
@@ -96,5 +98,342 @@ public void SPOptions_MetadataCacheDuration_DefaultValue()
...
+        [TestMethod]
+        public void SPOptions_MetadataCertificates_BothBecomesSigningWhenFuturePublished()
+        {
+            var subject = new SPOptions();
+            subject.ServiceCertificates.Add(new ServiceCertificate { Certificate = SignedXmlHelper.TestCert });
+            subject.ServiceCertificates.Add(new ServiceCertificate { Status = CertificateStatus.Future, Certificate = SignedXmlHelper.TestCert2 });
+
+            var result = subject.MetadataCertificates;
+            result.Count.Should().Be(2);
+            result[0].Use.Should().NotBe(result[1].Use);
```
AndersAbel: The name of the tests is "BothBecomesSigning" but this indicates they're expected to be different. Is the naming wrong? Or the assertion?

explunit: These two are configured as `Both`, but when one of them is Status=Future, then the Status=Current one becomes `Signing`. Perhaps the assertions should be more explicit.  I tried to avoid assuming any order of publishing but I could do a lookup by serial number and check how it's published.

Kentor.AuthServices/Configuration/SPOptions.cs
```diff
@@ -217,9 +220,107 @@ public IdentityConfiguration SystemIdentityModelIdentityConfiguration
...
+        }
+
+        /// <summary>
+        /// Certificate for use in signing outbound requests
+        /// </summary>
+        public X509Certificate2 SigningServiceCertificate
+        {
+            get
+            {
+                var decryptionCertificates = ServiceCertificates
```
AndersAbel: Shouldn't the variable be named signingCertificates?

explunit: This is for use in signing outbound requests, so I think it must resolve to just one. For metadata, or course, there might be multiple.

Kentor.AuthServices/Configuration/SPOptions.cs
```diff
@@ -217,9 +220,107 @@ public IdentityConfiguration SystemIdentityModelIdentityConfiguration
...
-        /// Certificate for service provider to use when decrypting assertions
+        /// Certificates to be published in metadata
         /// </summary>
-        public X509Certificate2 ServiceCertificate { get; set; }
+        public ReadOnlyCollection<ServiceCertificate> MetadataCertificates
+        {
+            get
+            {
+                var futureEncryptionCertExists = publishableServiceCertificates
+                    .Any(c => c.Status == CertificateStatus.Future && c.Use == CertificateUse.Encryption);
```
AndersAbel: I think this should be `c.Status == CertificateSTatus.Future && (c.Use == CertificateUse.Encryption || CertificateUse.Both))`

Kentor.AuthServices/Configuration/SPOptions.cs
```diff
@@ -217,9 +220,107 @@ public IdentityConfiguration SystemIdentityModelIdentityConfiguration
...
+                    .Any(c => c.Status == CertificateStatus.Future && c.Use == CertificateUse.Both);
+
+                foreach(var cert in metaDataCertificates)
+                {
+                    // Just like we stop publishing Encryption cert immediately when a Future one is added,
+                    // in the case of a "Both" cert we should switch the current use to Signing so that Idp's stop sending
+                    // us certs encrypted with the old key
+                    if (cert.Use == CertificateUse.Both && cert.Status == CertificateStatus.Current && futureBothCertExists)
+                    {
+                        cert.Use = CertificateUse.Signing;
```
AndersAbel: Isn't `cert` only one instance? Changing the Use of that will affect other places using the cert too. It will cause the certificate to no longer be used for decryption. It can be fixed by changing the query on line 284 to Select a new instance of ServiceCertificate, just copying over all properties.

Kentor.AuthServices/Configuration/SPOptions.cs
```diff
@@ -217,9 +220,107 @@ public IdentityConfiguration SystemIdentityModelIdentityConfiguration
...
+                    }
+
+                    if (cert.MetadataPublishOverride == MetadataPublishOverrideType.PublishEncryption)
+                    {
+                        cert.Use = CertificateUse.Encryption;
```
AndersAbel: Same issue here, doing a copy of the cert object in the query will solve this too.

Kentor.AuthServices.Tests/Configuration/SPOptionsTests.cs
AndersAbel: Now I see. I read "Both" in the test title as "Both Certificates", not as "Use=Both". Might be worth a short comment to avoid confusing someone else reading the code later.

Kentor.AuthServices/Configuration/SPOptions.cs
AndersAbel: I'm looking at the `decryptionCertificates` variable - looks like code is copy pasted from property above. The conditions are changed, but not the var name.

Kentor.AuthServices.Tests/Configuration/SPOptionsTests.cs
explunit: Updated the name of the test, added a comment, and made the assertions more explicit about what they're checking.

Kentor.AuthServices/Configuration/SPOptions.cs
explunit: I see now. Fixed.

Kentor.AuthServices/Configuration/SPOptions.cs
explunit: You're right. Fixed and added test `SPOptions_MetadataCertificates_OnlyOneEncryptionPublished_FutureBoth`

Kentor.AuthServices/Configuration/SPOptions.cs
explunit: Fixed and added lines to several existing tests to verify that the cert in the original collection is not modified.

albinsunnanbo:
Regarding the breaking changes in the upcoming version, would it be possible to reintroduce the readme-file and make it pop up once you update the NuGet package to notify of the breaking changes?

AndersAbel:
@albinsunnanbo I added it as an issue #382, but didn't have time to include it in the 0.15.0 release. I tried to be very clear in the release notes about breaking changes though.

cornem:
@AndersAbel I noticed that you guys did some work on this. I'm wondering if this means you now support signed HTTP POST AuthNRequests.

About a year ago I forked your project and [adding signing and POST support](https://github.com/cornem/authservices/commit/b5fc809fa231745adc70950a2869377ad96c2158), which worked fine for our specific needs.

Can I now achieve the same with the latest NuGet packages and how would I configure it?

AndersAbel:
@cornem, Yes 0.15.1 supports signed AuthnRequests over HTTP POST.

You need to add a serviceCertificate in the config and also set spOptions.AuthenticateRequestSigningBehavior = SigningBehavior.Always.

cornem:
@AndersAbel I had just figured it out. I'm not using a config file, so instead I have done this:

    var spOptions = new SPOptions
    {
        EntityId = new EntityId(new Uri(host, "AuthServices").AbsoluteUri),
        ReturnUrl = new Uri(host, "Account/ExternalLoginCallback"),
        Organization = organization,
        AuthenticateRequestSigningBehavior = SigningBehavior.Always
    };

    spOptions.ServiceCertificates.Add(
        new ServiceCertificate
        {
            Certificate = new X509Certificate2(HostingEnvironment.MapPath("path/to/cert.pfx"), "secret"),
            Status = CertificateStatus.Current,
            Use = CertificateUse.Signing
        }
    );

I think the request is correctly signed, and sent to the IdP, but then I get `No Saml2 Response found in the http request.` What could be the reasong for this?

(If unrelated, maybe I should create a new issue.)

AndersAbel:
@cornem Please open another issue for the "No Saml2 Response found" exception.

Can you provide some details on the response sent from the Idp by looking in the browser's dev tools? If possible I'd like to see http method, url and request body.

PR #405, opened 2016-02-11
Signing and multi instance okta

dahlsailrunner:
Here's a PR for the latest Okta documentation changes, which include the signing request functionality (see top level step 3 for the change), and a new step 6 to describe what we discussed regarding setting up support for multiple Okta instances.  

Let me know if you want me to make any edits or if something looks amiss.

Cheers!  :)


AndersAbel:
Thanks for excellent documentation. I made a small change: I removed the paragraph about adding an empty System.IdentityModel section to web config. When using code based config (passing false to the `KentorAuthServiceOptions` ctor) it is no longer needed.

Issue #408, opened 2016-02-16
Ignore SPSSODescriptor on metadata parsing


AndersAbel:
There's been issues with the parsing the InCommon metadata (#386), one of them is due to SPs having duplicate index numbers on their AssertionConsumerServices. Since we're not interested in the other SPs anyways we could just skip over them when parsing metadata.

Duplicate indexes is actually a violation of the SAML standard:
>A required attribute that assigns a unique integer value to the endpoint so that it can be referenced in a protocol message. The index value need only be unique within a collection of like elements contained within the same parent element (i.e., they need not be unique across the entire instance).

But if other implementations can handle duplicate indexes it looks like we're asking for too much trouble if we start blocking when they are found.

Override the `ReadServiceProviderSingleSignOnDescriptor()` method in `ExtendedMetadataSerializer` and loop the reader until an end SPSSODescriptor element is found.

jpsullivan:
I've begun working on this. So far, I have an implementation for the `ReadServiceProviderSingleSignOnDescriptor()` method, but I'm now seeing that adding an Organization URL is now throwing the same error within the `ReadOrganization()` method. I'd like to avoid copying the entire method and adding a check when inserting new organization URL's, but I'm not entirely sure what the alternative would be.

Issue #169, opened 2014-12-04
Multiple OrganizationDisplayName in federation metadata


firegrass:
The following metadata causes the System.Identity to exception.

```
    <Organization>
      <OrganizationName xml:lang="en">NHS England</OrganizationName>
      <OrganizationDisplayName xml:lang="en">NHS England</OrganizationDisplayName>
      <OrganizationDisplayName xml:lang="en">North West</OrganizationDisplayName>
      <OrganizationDisplayName xml:lang="en">North East of England</OrganizationDisplayName>
```
See https://fed.openathens.net/oafed/metadata?entityID=https://idp.eng.nhs.uk/openathens&browser

I realise that this issue is out of the control of this library, however I am wondering if either a workaround by overriding ReadOrganization method in ExtendedMetadataSerializer is appropriate OR is this non standards compliant and should be fixed in the metadata?

PR #409, opened 2016-02-16
Ignore SPSSODescriptor when parsing metadata

jpsullivan:
This PR is in reference to #408.

Includes two fixes, one which completely ignores any information
contained in the `SPSSODescriptor` element, and the other for how
`OrganizationUrl`s are stored.

InCommon seems to have multiple IdP's which contain multiple
organization URL's of the same language. This caused another duplicate
key exception. Added a check before inserting to the URL list that the
specified language doesn't already exist.

### DIFF
--- a/Kentor.AuthServices.Tests/ExtendedMetadataSerializerTests.cs
+++ b/Kentor.AuthServices.Tests/ExtendedMetadataSerializerTests.cs
@@ -301,33 +301,9 @@ gosrSG6sO3IPeL4BncKqqZO2FokfZbaqPBv6xmoKsVTUTQRfNEks84dRiG0MjqBncR+B6CIrCv2a
 
             var organizationInfo = entityDescriptor.Organization;
 
-            organizationInfo.Names[0].Name.Should().Be("Test Org");
-            organizationInfo.DisplayNames[0].Name.Should().Be("Test Org Name");
-            organizationInfo.Urls.Count.Should().Be(2);
-        }
-
-        [TestMethod]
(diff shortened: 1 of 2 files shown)


AndersAbel:
I've seen organization problems before, see #169. Great that you address that too. I see you've tried to keep parsing the Organization data, I don't know if that's really necessary. Right now that data is ignored when adding the Idps anyway. Or do you think that the Organization data should be added to the `IdentityProvider` class? If you want to build some kind of UI to choose IDP (i.e. embedded discovery) I assume you'd want the metadata.

There should be also be a couple of tests for the MetadataLoader would be good to have to show the behaviour. One could be a simple EntityDescriptor with some rubbish in the SPSSODescriptor element. Another test could be with multiple organization elements.

jpsullivan:
> If you want to build some kind of UI to choose IDP (i.e. embedded discovery) I assume you'd want the metadata.

That was my main reason for preserving the rest of the organization data. It's not particularly needed for our use case, but I figured others may want to preserve that.

As for tests, I totally agree. I'll add a few later on today.

jpsullivan:
I'm not entirely sure why I have the version bump commits in my history. I'd be happy to get rid of them, but unfortunately my git knowledge is somewhat exhausted past basic PR contributions.

AndersAbel:
Thanks @jpsullivan, code looks like nice and clean and there's test coverage. Everything I can ask for. When do you need this in an official release? Can you wait a couple of weeks?

jpsullivan:
No problem. Right now I'm just bin deploying a local copy which seems to work fine.

PR #424, opened 2016-03-06
423

federicobarera:
Implemented a post redirect logout logic compatible with Idsv3.
The implementation does not require code changes in the core library


AndersAbel:
Looks interesting, but I don't want know if I want another project added to the solution, or if I want it in the existing project, or if all of this really should be done in IdentityServer3 instead. I've talked to @brockallen about some updates to the IdSrv3 logout flow, but haven't had time to look into it yet.

federicobarera:
No problemo. You can close the pull, I just wanted to highlight the changes.

I've seen the ticket in idsv but looks like it's going to take a while before a solution.
In the meantime I've found an easier way. Hopefully gonna help before an implementation is done in one of the two libraries. 


`


            app.Use(async (context, next) =>
            {
                //save idsv logout state if present. this has to be done before headers are written
                var signOutMessageId = context.Environment.GetSignOutMessageId();
                if (signOutMessageId != null)
                    context.Response.Cookies.Append("state", signOutMessageId);

                var state = context.Request.Cookies["state"];
                if (state != null && context.Response.StatusCode == 200 && context.Request.Uri.ToString() == $"{ConfigurationHelper.Authority}logout")
                {
                    context.Response.Redirect($"{ConfigurationHelper.Authority}signoutcallback");
                    return;
                }

                await next.Invoke();
            });`

PR #429, opened 2016-03-23
Service Certificate validation

### LINKED ISSUE #412: CheckSignature with SHA256
I have a ServiceCertificate which is the StubIdp you provide for test, an Identity Provider with a signing SHA256 certificate, and I added the Options.GlobalEnableSha256XmlSignatures(). In the zip in attachments, you can find an example of a samlp response from my IDP.
Unfortunately, I can't pass through the CheckSignature, and I have a CryptographicException in the Saml2Response.retrieveAssertionElements method. 
To be sure, I tried to decrypt the response with the private key of the StubIdp certificate and it works fine (https://www.samltool.com/attributes.php). 
Is there something I miss?

Thanks for your help !

[test_serviceprovider.zip](https://github.com/KentorIT/authservices/files/141734/test_serviceprovider.zip)

explunit:
Fixes #412 by adding `ServiceCertificateCollection` so we can validate the certificates as they are added. This turned into a lot more changes than I would have expected for such a simple goal, and I welcome any feedback/criticism on this approach.

Note that we already had a `ServiceCertificateCollection` but it was really a collection of `ServiceCertificateElement`, so I renamed it as such.

We also had a `ServiceCertificateCollectionExtensions` (for a convenience `Add` method) and I've merged that into this new class as well.

As mentioned in #412, the only reliable method of validating the certificate is to try to encrypt/decrypt something with it. If I ever get a satisfactory answer to [this StackOverflow question](http://stackoverflow.com/questions/35816205/verify-certificate-is-valid-for-encryption-decryption) I can come back and enhance the approach. For now, the performance impact is not a large concern since it generally happens once at application startup.

The default Kentor.AuthServices.Tests.pfx file was not valid for use in decryption, so I renamed it to Kentor.AuthServices.TestsSignOnly.pfx and used the previous Kentor.AuthServices.Tests3.pfx as the new default. I had to adjust several tests and sample applications to match these new defaults, but I thought it was less disruptive than fixing everything that used the default certificate to *not* use the default `CertificateUse = Both`.


AndersAbel:
I was also under the impression that would be a small fix... I'll review it and get back.

Kentor.AuthServices/Configuration/ServiceCertificateCollection.cs
```diff
@@ -1,58 +1,72 @@
...
-        /// <summary>
-        /// Get a strongly typed enumerator.
-        /// </summary>
-        /// <returns>Strongly typed enumerator.</returns>
-        public new IEnumerator<ServiceCertificateElement> GetEnumerator()
-        {
-            return base.GetEnumerator().AsGeneric<ServiceCertificateElement>();
+            if (!item.Certificate.HasPrivateKey)
+            {
+                throw new ArgumentException(@"Provided certificate is not valid for encryption/decryption because it does not contain a private key.");
```
AndersAbel: I think we should simplify it to "Provided certificate is not valid because it does not contain a private key.". Without a key we can neither decrypt, nor verify a signature.

Also, it would be good if the exception message contained the thumb print of the cert, to make it easy to find what cert is wrong in case there are multiple certificates in the configuration.

Kentor.AuthServices/Configuration/ServiceCertificateCollection.cs
```diff
@@ -1,58 +1,72 @@
...
+            }
+
+            if (item.Use == CertificateUse.Encryption || item.Use == CertificateUse.Both)
+            {
+                if (!CertificateWorksForDecryption(item.Certificate))
+                {
+                    throw new ArgumentException(@"Provided certificate is not valid for encryption/decryption. It can only be used for other purposes such as signing.");
```
AndersAbel: Is it possible to be a bit more clear on what to do? Maybe add something like "if you only want to use the certificate for signing, you should add a Usage=Signing config".

Kentor.AuthServices/Configuration/ServiceCertificateCollection.cs
explunit: Good point. I'll fix it. We need the private key regardless of which `Use`. I just forgot the change the message when I moved it.

explunit: Updated

Kentor.AuthServices/Configuration/ServiceCertificateCollection.cs
explunit: Updated. Let me know if it's more clear.

AndersAbel:
I've been thinking about where to best do the validation. The current approach which makes it possible to carry invalid certificates over into the decryption/signing validation code doesn't feel good. It would be better to do the validation on configuration creation and then being able to trust it. With the `ISPOptions` interface being free to implement on your own, it is possible to get past the validation in the collection, so that's what I think is the problematic point.

What if the validation instead was moved to the `ServiceCertificate` class? Making `Usage` and `Certificate` readonly and provide a ctor that sets them. Then the ctor could do the validation and there would be no need for validation when we get to decryption/signature validation.

explunit:
@AndersAbel Rebased this branch. Let me know of any remaining issues.

PR #439, opened 2016-04-01
Add port in destination url

dufourpy:
Here is the modification with a test from a branch from master of my own repo

Thanks


AndersAbel:
Thanks for opening another PR. The code coverage fails though, probably because you've used a `?.` operator where there is no need for.

PR #443, opened 2016-04-04
442 Saml2StatusCode Mapping completed

federicobarera:


AndersAbel:
Thanks, great that you included the explaining text from the spec in the XML comments.

PR #450, opened 2016-04-15
Support for SAML2 Scoping

jimmytoenners:
This pull request introduces support for the Saml2 Scoping element. The corresponding Saml2P classes have been added and the IdentityProvider class has been extended with a Saml2ScopingProvider property. For the use case where scoping is set once per identity provider a specialized provider (SingleSaml2ScopingProvider) has been introduced. For the scenarios where dynamic scoping is needed per request an extension point has been introduced in SignInCommand that makes a call to the scoping provider on the identity provider (immediately following the call to CreateAuthenticateRequest) passing the relay state dictionary and the unsigned authentication request. 

Our use case is integration with a multitenant identity provider (WAYF.dk) where the Scoping element is used to send the user directly to the organization he belongs to. The scoping element we send depend on which customer subdomain the user accesses.


jimmytoenners:
Do you know what goes wrong in the build? seems master also fails?

AndersAbel:
@jimmytoenners Looks like my commit earlier today broke the test runs - they're not run at all. No idea why, I'll have to look into it. I cancelled your build - it was obviously stalled. The build usually completes under 2 minutes, so something is wrong.

AndersAbel:
@jimmytoenners I retried both the master build and your branch now on Appveyor and they're now passing. Looks like it was some temporary error on the CI server.

Kentor.AuthServices/SAML2P/Saml2RequesterId.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <param name="id">The identifier.</param>
+        public Saml2RequesterId(Uri id)
+        {
+            Id = id;
+        }
+
+        /// <summary>
+        /// Gets and sets the id.
+        /// </summary>
+        public Uri Id { get; set; }
```
AndersAbel: Is this an Entity ID? In that case, please use the EntityId class instead of URI.

AndersAbel:
The code looks clean and follows the overall principles. There are a few minor things, but I'll either update those myself when merging or take it later when we have finalized the overall structure.

I've been thinking a bit of how to handle the callback and the best way to do that. Some thoughts I have (in random order):
* Using an `Action<>` or `Func<>` instead of an interface feels more light weight and modern.
* Putting the callback on the `IdentityProvider` makes it hard to use the feature when the Idp is dynamically created by a federation.
* Relying on the relayData to pass information back to the application works for the Owin middleware, but not for the httpmodule and the Mvc controller. I think that it will also work with a planned future ASP.NET Core Middleware. I think that this is fine - there is no good mechanism to pass parameters in the http module and MVC controller. That would require introduction of a new API where the application directly calls a method instead of redirecting to a URL.

Based on that, I'd like to hear your thoughts on the following changes:
* Add a new property `IOptions.Notifications` of type `KentorAuthServicesNotifications` - modelled in the same way as the other Katana middlewares  [e.g. WsFed](https://katanaproject.codeplex.com/SourceControl/latest#src/Microsoft.Owin.Security.WsFederation/WsFederationAuthenticationNotifications.cs).
* Add an `Action<Saml2AuthenticationRequest, IdentityProvider, IDictionary<string, string>> AuthenticateRequestCreated` to `KentorAuthServicesNotifications` and call it from the same place in the `SignInCommand` (after creation of message, but before encoding it through the binding).

The reason I suggest those changes are:
* More in line with how the Katana Middleware works.
* Sets a good and light weight foundation for further extensions points (which will be required).

So, what do you think? Does my ramblings make any sense?

jimmytoenners:
Yes I have been thinking along the same lines and actually think you are right that a generic AuthenticationRequestCreated notification would be a cleaner and more generic way to handle this scenario. Are you going to give it a shot or should we?

Regarding the use of a new type for RequesterId I guess I just did that because it is just anyURI in the schema. Also note that the schema dictates a sequence which means the elements should be in correct order (they are now). While we are at it we could add the remaining properties and make it complete. But this could also be work another PR takes care of I guess. 

Schema below for reference: 

    <element name="AuthnRequest" type="samlp:AuthnRequestType"/>
    <complexType name="AuthnRequestType">
        <complexContent>
            <extension base="samlp:RequestAbstractType">
                <sequence>
                    <element ref="saml:Subject" minOccurs="0"/>
                    <element ref="samlp:NameIDPolicy" minOccurs="0"/>
                    <element ref="saml:Conditions" minOccurs="0"/>
                    <element ref="samlp:RequestedAuthnContext" minOccurs="0"/>
                    <element ref="samlp:Scoping" minOccurs="0"/>
                </sequence>
                <attribute name="ForceAuthn" type="boolean" use="optional"/>
                <attribute name="IsPassive" type="boolean" use="optional"/>
                <attribute name="ProtocolBinding" type="anyURI" use="optional"/>
                <attribute name="AssertionConsumerServiceIndex" type="unsignedShort" use="optional"/>
                <attribute name="AssertionConsumerServiceURL" type="anyURI" use="optional"/>
                <attribute name="AttributeConsumingServiceIndex" type="unsignedShort" use="optional"/>
                <attribute name="ProviderName" type="string" use="optional"/>
            </extension>
        </complexContent>
    </complexType>
    <element name="NameIDPolicy" type="samlp:NameIDPolicyType"/>
    <complexType name="NameIDPolicyType">
        <attribute name="Format" type="anyURI" use="optional"/>
        <attribute name="SPNameQualifier" type="string" use="optional"/>
        <attribute name="AllowCreate" type="boolean" use="optional"/>
    </complexType>
    <element name="Scoping" type="samlp:ScopingType"/>
    <complexType name="ScopingType">
        <sequence>
            <element ref="samlp:IDPList" minOccurs="0"/>
**            <element ref="samlp:RequesterID" minOccurs="0" maxOccurs="unbounded"/>
**        </sequence>
        <attribute name="ProxyCount" type="nonNegativeInteger" use="optional"/>
    </complexType>
   ** <element name="RequesterID" type="anyURI"/>
   ** <element name="IDPList" type="samlp:IDPListType"/>
    <complexType name="IDPListType">
        <sequence>
            <element ref="samlp:IDPEntry" maxOccurs="unbounded"/>
            <element ref="samlp:GetComplete" minOccurs="0"/>
        </sequence>
    </complexType>
    <element name="IDPEntry" type="samlp:IDPEntryType"/>
    <complexType name="IDPEntryType">
        <attribute name="ProviderID" type="anyURI" use="required"/>
        <attribute name="Name" type="string" use="optional"/>
        <attribute name="Loc" type="anyURI" use="optional"/>
    </complexType>
    <element name="GetComplete" type="anyURI"/>

AndersAbel:
I can give the events a shot and then merge your changes on top of them. I think I have time to do that tomorrow or on Wednesday.

I don't think that there's any need right now to implement the rest of the optional Scoping fields. They can be added when needed, as long as the current class structure allows that (and from what I've seen it does).

jimmytoenners:
I have implemented the event mechanism. Please have a look at the code in this branch.

AndersAbel:
@explunit you've done some config works so I'd like to hear if you have any opinion on having the notifications directly on `IOptions` instead of on `SPOptions`. They're not really part of the SP behaviour so I thought putting them directly on `IOptions` made sense.

AndersAbel:
@jimmytoenners Thanks, I'll have a look and get back.

explunit:
Your logic regarding notifications on `IOptions` makes sense to me.

AndersAbel:
Doing reviews and working with the code I have to change where the `Notifications` goes. They have to be in `SPOptions` because that's what's passed around in many places. Having the `Notifications` in `IOptions` would mean that the `IdentityProvider` ctor would need to be changed. That's a too big breaking change.

AndersAbel:
I've now reviewed this and made some updates. It's available in the [Scoping](https://github.com/KentorIT/authservices/compare/Scoping?expand=1) branch. Before merging, I'd like to have your comments if I've destroyed something important. What I've done:

* Some general fixes to follow coding style (delete commented out code, adding missing XML Comments etc.)
* Removed the ISaml2ScopingProvider interface - it's up to the user of the library to do whatever work needed in the notification callback.
* Kept the `Notifications` in `IOptions` - I was wrong, the entire `IOptions` are available where needed.
* Only using Ctor arguments for required stuff, the rest can be set through properties.
* Using Uri and EntityId types instead of strings where appropriate.

jimmytoenners:
Please introduce this in Saml2IdPEntry.cs:  

idpEntryElement.AddAttributeIfNotNullOrEmpty("Loc", Location?.OriginalString);

And this test in Saml2P\Saml2IdpEntryTests.cs
        [TestMethod]
        public void Saml2IdpEntry_LocationIsOptional()
        {
            var subject = new Saml2IdpEntry(new EntityId("urn:foo")) {Location = null };

            subject.Invoking(s => s.ToXElement())
                .ShouldNotThrow();
        }

For reference:
<complexType name="IDPEntryType">
    <attribute name="ProviderID" type="anyURI" use="required"/>
    <attribute name="Name" type="string" use="optional"/>
    <attribute name="Loc" type="anyURI" use="optional"/>
</complexType>

AndersAbel:
@jimmytoenners Thanks for noticing the bug when Location is null. I'll add that fix+test and then merge.

PR #454, opened 2016-04-18
Correct namepace of Artifact

SWSSolutions:
Correct namepace of Artifact (samlp:Artifact instead of saml:Artifact) see: https://en.wikipedia.org/wiki/SAML_2.0#Artifact_Resolution_Protocol


AndersAbel:
Thanks for noticing, this should of course be fixed. Before merging the tests need to be updated too. I'll look into that (or if you have time to fix the broken test, that's of course welcome).

SWSSolutions:
Thank you for your message, unit tests are updated.

PR #462, opened 2016-05-01
Added SubjectConfirmationData

### LINKED ISSUE #461: SubjectConfirmation missing SubjectConfirmationData
Hi,

I've noticed that the SAML response tokens generated by the StubIdP, have a SubjectConfirmation with a method of `urn:oasis:names:tc:SAML:2.0:cm:bearer`, however they're missing a SubjectConfirmationData child element. I'm aware that you're providing a Conditions element to accompany it, but I believe this is still in conflict with the [SAML specification](http://docs.oasis-open.org/imi/identity/cs/imi-saml2.0-profile-cs-01.html#_Ref248548678). The reference in particular is 

> In the case of bearer assertions, the `<saml:SubjectConfirmation>` element **MUST** include a `<saml:SubjectConfirmationData>` element containing a NotOnOrAfter XML attribute to limit their use, typically to a very short window of time, although the exact duration may be use case dependent.

Am I misreading something?

Also, all response tokens generated by the StubIdp are failing to [pass validation](https://www.samltool.com/validate_response.php).

I came across this when attempting to integrate SSO with Tableau (Which are using Spring Framework) is rejecting responses with an error 

> Assertion invalidated by subject confirmation - can't be confirmed by the bearer method

Any insight would be appreciated.

Thanks,

Trent

tgardner:
Fixes #461 

This pull request adds SubjectConfirmationData to the saml response token as defined in [the specification](http://docs.oasis-open.org/imi/identity/cs/imi-saml2.0-profile-cs-01.html#_Ref248548678).
The missing confirmation data was causing SAML tokens to fail validation.

### DIFF
--- a/Kentor.AuthServices.Tests/Saml2AssertionExtensionsTests.cs
+++ b/Kentor.AuthServices.Tests/Saml2AssertionExtensionsTests.cs
@@ -73,6 +73,35 @@ namespace Kentor.AuthServices.Tests
         }
 
         [TestMethod]
+        public void Saml2AssertionExtensions_ToXElement_Subject_ConfirmationData()
+        {
+            var subjectName = "JohnDoe";
+            var destination = new Uri("http://sp.example.com");
+            var inResponseTo = new Saml2Id("abc123");
+
+            var assertion = new Saml2Assertion(
+                new Saml2NameIdentifier("http://idp.example.com"))
+            {
+                Subject = new Saml2Subject(new Saml2NameIdentifier(subjectName))
+            };
+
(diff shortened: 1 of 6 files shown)


tgardner:
Hi @AndersAbel, I have updated this to use the SubjectConfirmation object within the Saml2Subject rather than hardcoding the confirmation data, seemed like a cleaner solution

AndersAbel:
Looks good with the update using the `SubjectConfirmation` object. I'll merge it.

Thanks for contributing.

Issue #477, opened 2016-05-19
KeyNotFoundException in case of omitted optional Issuer tag


pshirshov:
In case when /Response/Issuer attribute is missing, Kentor throws KeyNotFoundException.

This response works:

```
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="_8e8dc5f69a98cc4c1ff3427e5ce34606fd672f91e6" Version="2.0" IssueInstant="2014-07-17T01:01:48Z" Destination="http://sp.example.com/demo1/index.php?acs" InResponseTo="ONELOGIN_4fee3b046395c4e751011e97f8900b5273d56685">
  <saml:Issuer>http://idp.example.com/metadata.php</saml:Issuer>
  <samlp:Status>
    <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/>
  </samlp:Status>
  <saml:Assertion xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" ID="_d71a3a8e9fcc45c9e9d248ef7049393fc8f04e5f75" Version="2.0" IssueInstant="2014-07-17T01:01:48Z">
    <saml:Issuer>http://idp.example.com/metadata.php</saml:Issuer>
```

This one not:

```
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="_8e8dc5f69a98cc4c1ff3427e5ce34606fd672f91e6" Version="2.0" IssueInstant="2014-07-17T01:01:48Z" Destination="http://sp.example.com/demo1/index.php?acs" InResponseTo="ONELOGIN_4fee3b046395c4e751011e97f8900b5273d56685">
  <samlp:Status>
    <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/>
  </samlp:Status>
  <saml:Assertion xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" ID="_d71a3a8e9fcc45c9e9d248ef7049393fc8f04e5f75" Version="2.0" IssueInstant="2014-07-17T01:01:48Z">
[...]

PR #478, opened 2016-05-19
Issuer extraction fixed

FeiWongReed:
Fix for #477

### DIFF
--- a/Kentor.AuthServices/SAML2P/Saml2Response.cs
+++ b/Kentor.AuthServices/SAML2P/Saml2Response.cs
@@ -99,7 +99,16 @@ namespace Kentor.AuthServices.Saml2P
                 secondLevelStatus = xml["Status", Saml2Namespaces.Saml2PName]["StatusCode", Saml2Namespaces.Saml2PName]["StatusCode", Saml2Namespaces.Saml2PName].Attributes["Value"].Value;
             }
 
-            Issuer = new EntityId(xmlElement["Issuer", Saml2Namespaces.Saml2Name].GetTrimmedTextIfNotNull());
+            var maybeIssuer = xmlElement["Issuer", Saml2Namespaces.Saml2Name].GetTrimmedTextIfNotNull();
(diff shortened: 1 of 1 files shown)


explunit:
Does the standard say that the first assertion is authoritative for issuer if it's not found in the top-level response tag? We may want to be cautious about this, given the signature-wrapping attacks. It would seem that if we adopt your suggested fix that the signature checking logic would need to change to lookup issuer on every assertion rather than using this top-level issuer property to find the Ipd & its keys.

Also, on PR logistics, some things to note:
 * You'll need some unit tests to fix the code coverage failure you see above
 * You should do the PR from a branch in your repo rather than master. 
See https://github.com/KentorIT/authservices/blob/master/CONTRIBUTING.md

AndersAbel:
@explunit is right about the signature processing being affected. This is what the WebSSO Profile section 4.1.4.2 says:

> If the <Response> message is signed or if an enclosed assertion is encrypted, then the <Issuer> element MUST be present. Otherwise it MAY be omitted.

So if there is no Issuer on the Response, the `Saml2Response.Issuer` property should be left as null. The signature validation code need to be adjusted accordingly so that any signed assertions are validated against the key of the `issuer` of the assertion. I think that there are cases with proxying where an idp might send a response containing assertions issues by another idp.

AndersAbel:
Closing due to inactivity. If you're still working on this, please leave a comment and I'll reopen.

FeiWongReed:
I've reported a bug and proposed quick fix. Ok, code-style-coverage-and-so-on make my fix unacceptable, but I have no time to do everything in perfect way, I'm sorry. Maybe later, not today.

AndersAbel:
I understand if you don't have time to work further on it, but quick fixes that do not stand up to the quality standards of the library will not be accepted. In this case it was not only test coverage that lacked but also logic around the signature validation. That is a **very** sensitive part of the code and changes must be very well worked through to be accepted.

Issue #486, opened 2016-06-01
Start application without registered fail


maxim-s:
if start application using OWIN middleware without any registered IDPs, will get error 
```
Sequence contains no elements

Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.InvalidOperationException: Sequence contains no elements

Source Error: 


Line 113:                lock(dictionary)
Line 114:                {
Line 115:                    return dictionary.Values.Skip(i).First();
Line 116:                }
Line 117:            }
```

If you can dynamic switch between IDPs, and also can dynamic add IDPs, why it's impossible to start without any registered IDP ?

AndersAbel:
Because we never thought about scenario. Can you provide a more complete stack trace? Do you have any suggestions on what to do instead?

maxim-s:
Actually I thing it's not necessary to run SignInCommand if we have null idp

```
Exception Details: System.InvalidOperationException: Sequence contains no elements

Source Error: 

An unhandled exception was generated during the execution of the current web request. Information regarding the origin and location of the exception can be identified using the exception stack trace below.

Stack Trace: 


[InvalidOperationException: Sequence contains no elements]
   System.Linq.Enumerable.First(IEnumerable`1 source) +264
   Kentor.AuthServices.Configuration.IdentityProviderDictionary.get_Item(Int32 i) +126
   Kentor.AuthServices.WebSso.SignInCommand.Run(EntityId idpEntityId, String returnPath, HttpRequestData request, IOptions options, IDictionary`2 relayData) +154
   Kentor.AuthServices.Owin.<ApplyResponseChallengeAsync>d__1.MoveNext() +511
   System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task) +92
   System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) +58
   Microsoft.Owin.Security.Infrastructure.<ApplyResponseCoreAsync>d__b.MoveNext() +292
   System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task) +92
   System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) +58
   Microsoft.Owin.Security.Infrastructure.<ApplyResponseAsync>d__8.MoveNext() +278
[...]

PR #487, opened 2016-06-01
#486 prevent sending Signup Command if no identity providers

maxim-s:


maxim-s:
looks like need to rerun  coverage/coveralls after PR synchronization , because when I added test still have the same coverage

AndersAbel:
The coverage is automatically recalculated after a rebuild. The [coveralls page](https://coveralls.io/github/KentorIT/authservices) shows that there's been two builds from your PR.

The non-covered condition in the if-statement is complex, so probably not all of those conditions are properly covered any more. Are they even needed? Might be that some of those null checks can be removed. Can Options.IdentityProviders really be null? If it can, there must be a test case that covers it.

maxim-s:
@AndersAbel yes sure,thank you, it's impossible to have null for IdentityProviders property

AndersAbel:
What about behaviour when AuthServices is explicitly stated in the AuthenticationTypes in the challenge? There are three cases. The wildcard case only makes sense when the middleware is set as active, but the reasoning should be the same.

* No AuthenticationTypes specified => Don't call SignInCommand if there is no Idp.
* One AuthenticationType specified and it matches Options.AuthenticationType => I prefer an exception to be thrown if there is no Idp. Otherwise it will just become a very strange error.
* Multiple AuthenticationTypes specified and one matches Options.AuthenticationType => Advanced scenario that indicates that the user knows what's going on. Don't call SignInCommand to let other middlewares have a chance.

What do you think about those alternatives?

maxim-s:
yes, reasonable cases, just want to check it in current environment

AndersAbel:
Closing due to no updates after my comments 1st of June. Please comment to reopen if you want to carry this forward.

PR #489, opened 2016-06-01
ASP.NET Core Support

kspearrin:
This PR adds full support for ASP.NET Core RC2.

Included is the `Kentor.AuthServices.AspNetCore` library and a `SampleAspNetCoreApplication` sample project.

An outstanding issue involves the use of the `KentorAuthServicesUserClaimsPrincipalFactory` in the sample project. This is needed to augment the logout claims currently and probably isn't the correct solution for this library since it requires a dependency on ASP.NET Core Identity. You may want to add an `AuthenticationHandler` wrapper of some sort like was discussed in chat.

Finally, this PR does not include the needed tests to achieve proper coverage. That will still need to be added. Comment documentation could also be improved.


AndersAbel:
This work has now been moved to the aspnetcore branch in the repo, so I'm closing this PR.

AndersAbel:
Marking as invalid just to get it out of the way from the filter I use for things to bring into a new release.

Issue #250, opened 2015-06-22
Redirect URL from RelayState


AndersAbel:
When using Idp-initiated login, there is a convention that the Idp can specify the URL to redirect after successful authentication, by including that in the RelayState parameter. We should add support for that.

See http://saml.xml.org/wiki/idp-initiated-single-sign-on-post-binding

CryptonZylog:
Seconded!

How would you handle this in ASP.NET execution pipeline? Setting a parameter within BeginRequest (overridable) or just grabbing the ReturnUrl from Forms Authentication?

AndersAbel:
@CryptonZylog I don't understand your question. Using a Redirect URL from relay state would only be useful with IDP initiated login. Using SP initiated login, the IDP is required to relay back whatever the SP put in the relaystate, so in that case the IDP can't (ab)use the RelayState parameter for a target URL.

CryptonZylog:
In our case, suppose a user follows a link to a specific protected page, but they are unauthenticated. They get redirected to IdP, authenticate, then IdP redirects them back to the static URL configured for AuthServices. I haven't had a chance to verify this yet, but it does not seem like AuthServices supports redirecting them back to the original URL they were requesting. Please correct me if I'm wrong.

albinsunnanbo:
If you have SP initiated login, AuthServices should respect the returnUrl parameter on the SignIn command.
/AuthServices/SignIn?returnUrl=https://myApp.com/aProtectedPage/ should fix the redirect. With no returnUrl parameter AuthServices uses the configured returnUrl from the config file.

AndersAbel:
@CryptonZylog AuthServices already supports redirecting back to the URL that triggered the signin. The MVC and Owin sample applications in the repo demonstrates that. This issue is about IDP-initiated login with a direct link to a specific resource.

CryptonZylog:
Excellent, thanks for the clarification.

nahidf:
Any update on this?

AndersAbel:
@nahidf It's unfortunately not prioritized for us at Kentor the moment and as far as I know there's noone else working on it. If you need it you are welcome to submit a Pull Request (please check contribution guidelines first). Kentor also offers sponsored development where you pay for a specific feature (like this) to be completed. Please mail me at anders.abel@kentor.se if that's interesting.

chrisnott:
[...]

PR #505, opened 2016-06-17
Create a separate InResponseToNoRelayStateValidationException

### LINKED ISSUE #504: Suggestion: Further break down the Saml2ResponseFailedValidationException
In our application we need separate handling of the unexpected InResponseTo variant of Saml2ResponseFailedValidationException:s.
This happens from time to time when users use the back button in the browser back to the IDP and tries to resubmit the same ticket.
We want to detect exactly this case and gracefully direct our users to a suitable solution.

Would it be most suitable to create a new exception inheriting from the Saml2ResponseFailedValidationException or add a separate ErrorCode property on the exception to handle the nuances?

albinsunnanbo:
Create a separate InResponseToNoRelayStateValidationException to be able to detect the specific kind of replay detection from client code

Fixes #504

@AndersAbel any thoughts on the naming of the exception? I went for very long and specific.


AndersAbel:
Just looked through and found that maybe this should be detected at another level, when an expected RelayState is not found. Now the data is extracted from the cookie in the `HttpRequestData` constructor. I'd rather move that to the binding because now that ctor has to extract the RelayState - thus duplicating that logic - to find the cookie name.

It would be better to bring on the entire cookie collection in the `HttpRequestData` and instead let the binding do the relaystate extraction.

What remains though is the problem that there is de facto standard that for unsolicited responses the RelayState contains the redirect url. AuthServices doesn't support it yet (#250), but including such a RelayState shouldn't result in an error being thrown. And it's not until in Saml2Response that we know if there is an InResponseTo.

So in the of this rambling (I'm thinking as I'm writing) the check is at the right level/place in the code. But I'd rather rename the exception to `UnexpectedInResponseToException` as it will be thrown both if the RelayState is lost, the cookie is lost or if an unsolicited response contains an InResponseTo.

I've opened a separate ticket for the RelayState extraction: #507.

albinsunnanbo:
Lots of rambling...

I renamed the exception.

It wasn't obvious to me at first that unsolicited request with an InResponseTo could be the cause too.
Should we improve the exception message further to indicate that possible cause?

AndersAbel:
Yes, an improved exception message would be good that indicates an unexpected InResponseTo would be good.

PR #524, opened 2016-07-06
Handle OneTimeUse SAML2Reponse

pi3k14:
Could you please merge this PR, to allow SAML2Response with OneTimeUse attribute to be processed?

### DIFF
--- a/Kentor.AuthServices/SAML2P/Saml2Response.cs
+++ b/Kentor.AuthServices/SAML2P/Saml2Response.cs
@@ -508,10 +508,11 @@ namespace Kentor.AuthServices.Saml2P
                     var token = (Saml2SecurityToken)handler.ReadToken(reader);
                     handler.DetectReplayedToken(token);
 
-                    // Have checked for replayed token, don't need to validate OneTimeUse (Saml2SecurityTokenHandler doesn't support it either)
-                    // a TechNet article that discusses this related to AD-FS http://social.technet.microsoft.com/wiki/contents/articles/2994.ad-fs-2-0-id4149-the-saml2securitytoken-is-rejected-because-the-saml2-assertion-specifies-a-onetimeuse-condition.aspx
-                    token.Assertion.Conditions.OneTimeUse = false;
-
+                    if (token.Assertion.Conditions.OneTimeUse == true && options.SPOptions.Compatibility.AcceptOneTimeUseAssertions)
+                    {
+                        token.Assertion.Conditions.OneTimeUse = false;
+                    }
+                 
                     var validateAudience = options.SPOptions
                                             .Saml2PSecurityTokenHandler
                                             .SamlSecurityTokenRequirement
@@ -523,8 +524,8 @@ namespace Kentor.AuthServices.Saml2P
                     handler.ValidateConditions(token.Assertion.Conditions, validateAudience);
 
                     sessionNotOnOrAfter = DateTimeHelper.EarliestTime(sessionNotOnOrAfter,
-                    token.Assertion.Statements.OfType<Saml2AuthenticationStatement>()
-                        .SingleOrDefault()?.SessionNotOnOrAfter);
+                                                                      token.Assertion.Statements.OfType<Saml2AuthenticationStatement>()
+                                                                        .SingleOrDefault()?.SessionNotOnOrAfter);
 
                     yield return handler.CreateClaims(token);
                 }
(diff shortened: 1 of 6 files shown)


Kentor.AuthServices/SAML2P/Saml2Response.cs
```diff
@@ -505,12 +508,19 @@ private IEnumerable<ClaimsIdentity> CreateClaims(IOptions options)
                     var token = (Saml2SecurityToken)handler.ReadToken(reader);
                     handler.DetectReplayedToken(token);
 
+                    // Have checked for replayed token, don't need to validate OneTimeUse (Saml2SecurityTokenHandler doesn't support it either)
+                    if (token.Assertion.Conditions != null)
```
AndersAbel: The if statement is not needed, at least not with the existing tests. Try adding a test with a saml2 response without conditions and see if it makes the if statement needed. I think it's blocked somewhere earlier as the replay detection requires a NotOnOrAfter.

AndersAbel:
I've looked into OneTime use before and never really understood what it means in the context of WebSSO. Your argument that the replay detection is a kind of onetime use detection makes sense. But on the other hand the generated claims might be reused by the application. Do you have any pointers to documentation on how to interpret OneTimeUse with the WebSSO profile?

Also you have to include a test for the new functionality (i.e. a test with a response/assertion that has OneTimeUse set)

pi3k14:
I think the following 

> "To support the single use constraint, a relying party should maintain a cache of the assertions it has processed containing such a condition. Whenever an assertion with this condition is processed, the cache should be checked to ensure that the same assertion has not been previously received and processed by the relying party"

from chapter 2.5.1.5 Element <OneTimeUse> in Assertions and Protocols for the OASIS Security Assertion Markup Language (SAML) V2.0 justifies this logic.

AndersAbel:
Yes, that snippet from 2.5.1.5 covers the act of establishing a session. What I'm concerned with is if the application uses the claims generated further. Can that be deemed as additional, repeated use?

Section 2.5.1.5 also contains this:
> In general, relying parties may choose to retain assertions, or the information they contain in some other form, for reuse. The `<OneTimeUse>` condition element allows an authority to indicate that the information in the assertion is likely to change very soon and fresh information should be obtained for each use.

For example, an application using AuthServices might save the users name, email, phone number etc. and then use those for displaying user records long after the session has expired. That would not be in line with a `<OneTimeUse>` restriction. I've tried to find some authoritative information on `<OneTimeUse>` in relation to the WebSSO profile but haven't found any.

What do you think about making this a compatibility setting that must be enabled? That way the user will know that they may not cache/save the generated claims. It might also be an idea to add a specific claim OneTimeUse=true to signal to the application that an identity is only valid for establishing a session - not for saving.

What do you think?

pi3k14:
There is a comment on [this](http://stackoverflow.com/a/33667725/2294065) answer in StackOverflow that concludes with using OneTimeUse as replay prevention in WebSSO, but I haven't found any SAML Errata that confirms this.

Adding the acceptance of OneTimeUse tokens as a compability/option setting would probably be wise, but sending it as an extra claim I think would be confusing. Especially if you do some claim transformations, eg. when using IdentityServer/Kentor as a bridge. What claims is then for OneTimeUse?

AndersAbel:
You're right in that a specific claim is probably a bad idea. The ideal place would be in the properties dictionary of the claims. But those are not serialized by the owin/katana cookie middleware (I never checked the SessionAuthenticationModule) so it's of no use.

For now, adding a compatibility flag and then setting `Conditions.OneTimeUse` to false will do. Would you mind updating the PR with a compatibility flag (maybe called `AcceptOneTimeUseAssertions`)? There is a `Compatibility` class that you can add it too. For completeness it should also be added to the `CompatibilityElement` that drives the config files - and to the documentation.

pi3k14:
Think that should be it - a bit uncertain about the intention of the SPOptions_Constructor_LoadsConfig() test though. Ah, just discovered (for coverage test), will push a minor update.

pi3k14:
Any chance getting this merged into a release?

AndersAbel:
Yes, it's on my should-already-have-done-list. I think that last time I looked at it that I wanted another test (that I thought of adding myself) that would test that it doesn't accept OneTimeUse if the compatibility flag isn't set.

pi3k14:
There you go - or if you want the test to work when no config is set something must be changed in SPOptions_Constructor_LoadsConfig()

pi3k14:
Need PR #545 before this build correctly again.

AndersAbel:
Closing and reopening to cause a rebuild now that #545 is merged.

PR #545, opened 2016-08-17
Update SHA256 messages and tests for .Net 4.6.2

explunit:
Now that SHA256 support is built-in to .Net 4.6.2 this test will not throw errors anymore. Here's one option for resolving this.

 1. Moved this check into a different method so that it can be excluded from code coverage.
 1. When the exception **is** thrown, updated the message to say that upgrading to .Net 4.6.2 is an option for resolving the error.
 1. Fixed wrong cert for the test, which was masked by the earlier error


AndersAbel:
Thanks for the fix @explunit

PR #555, opened 2016-08-30
Add ability to sign assertion

ShannonKasperNoesis:
I came across a requirement where the assertion needed to be signed in addition to the response. The proposed change seems to successfully meet the requirement. I think ideally it should be optional but didn't see an easy way to do that. I'm open to any suggestions.


AndersAbel:
Signing of the assertion shouldn't be done in the binding, it's part of the message data so it should be done when the XML is generated. And it should be configurable. And there has to be tests.

ShannonKasperNoesis:
Thanks, I'll work on updating my pull request.

smurfpandey:
I have verified, and this is working for my same requirement.

PR #557, opened 2016-09-04
persist challenge state across discovery service call

wagich:
This PR adds code to persist the `AuthenticationProperties` set by IdentityServer when rendering the login page across the following call to the discovery service, solving #182.

It persists the existing state using the already existing `StoredRequestState` object and the `RelayState` query string parameter when constructing the DS request and then rehydrates the challenge properties before running the `SignInCommand` in `KentorAuthServicesAuthenticationHandler.ApplyResponseChallengeAsync()`.


wagich:
Yay! I've had a crash course in unit testing and code coverage but it finally looks ok – please forgive the many iterations...

AndersAbel:
@wagich Thanks for the update. I'm lagging behind with pull request reviews and will have a more thorough look when I can find the time needed.

Shloosh:
Any updates on this?

PR #587, opened 2016-11-12
Compatibility setting and notifications for alternate logout state persistence

### LINKED ISSUE #509: NullReferenceException if request state cookie not passed to logout command
If the `Kentor.<randomvalue>` cookie is lost somewhere in the logout process (e.g. redirected via different domain), a NullReferenceException is thrown during the processing of LogoutResponse from Idp:

```
[NullReferenceException: Object reference not set to an instance of an object.]
Kentor.AuthServices.WebSso.LogoutCommand.HandleResponse(UnbindResult unbindResult, HttpRequestData request) +66
Kentor.AuthServices.WebSso.LogoutCommand.Run(HttpRequestData request, String returnPath, IOptions options) +190
```

This is because `StoredRequestState` is null in this line of code:

```
            return new CommandResult()
            {
                HttpStatusCode = HttpStatusCode.SeeOther,
                Location = request.StoredRequestState.ReturnUrl,
                ClearCookieName = "Kentor." + unbindResult.RelayState
            };
```

At minimum, I think we should throw a more informative exception, but I'm wondering if we can fall back to the `returnPath` parameter in this case? I see the note that it's only used during the start of SP-initiated logout, but I can't think of a reason why it couldn't also be used as a fallback location if the cookie does not contain it.

explunit:
Fixes #509 by allowing the developer to disable the logout state cookies, while exposing the RelayState value so they can use an alternate approach (e.g. database lookup) for the `StoredRequestState`

### DIFF
--- a/Kentor.AuthServices.Owin/CommandResultExtensions.cs
+++ b/Kentor.AuthServices.Owin/CommandResultExtensions.cs
@@ -55,7 +55,7 @@ namespace Kentor.AuthServices.Owin
         {
             var serializedCookieData = commandResult.GetSerializedRequestState();
 
-            if (serializedCookieData != null)
+            if (serializedCookieData != null && !string.IsNullOrEmpty(commandResult.SetCookieName))
             {
                 var protectedData = HttpRequestData.ConvertBinaryData(
                         dataProtector.Protect(serializedCookieData));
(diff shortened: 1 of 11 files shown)


AndersAbel:
Go ahead and merge this. I think that this might open up for CSRF issues on the logout sequence, but I can't really see any other way to solve it in a multi domain setup. As it is a non default behaviour, I guess only very few will actually use this setting. So for me, it's fine to merge.

PR #603, opened 2016-12-03
Add short troubleshooting section

albinsunnanbo:
Added a short troubleshooting section with some handy links.
Is this helpful and in the right location?


AndersAbel:
Good idea. But please put it further up or it won't be found.

albinsunnanbo:
High enough?

PR #614, opened 2016-12-21
Add UnsolicitedAuthnResponseReceived notification

zxjon22:
Based on feedback in #613, this pull request adds support for a new `UnsolicitedAuthnResponseReceived ` notification.

The notification is triggered by the AcsCommand after processing the request but before the existing `AcsCommandResultCreated` notification.

The issue in #613 can then be worked around in the host application.


AndersAbel:
Is this notification really needed? Why not use the existing `AcsCommandResultCreated` notification if they're called from the same place?

zxjon22:
Ah my mistake.

I'd been looking at checking `request.StoredRequestState` which can't be done from the `AcsCommandResultCreated` notification but it's sufficient just to look at `Saml2Response.InResponseTo  == null`

PR #716, opened 2017-05-10
Support client certificate for artifact resolve #367

### LINKED ISSUE #367: Support client certificate for artifact resolve
Support TLS client certificate to be used for the SOAP binding of artifact resolve queries.

MartijnKooij:
As discussed I added the artifact resolve TLS certificate as a new config setting. The artifact resolve soap request will add the certificate to the actual request if present.


AndersAbel:
I see you're trying to get the test coverage up, which is good. But some parts of this feature might be very hard to test successfully. If you find something that you can't find a way to test, bring it up and we'll discuss if `[ExcludeFromCodeCoverage]` is ok in those cases.

MartijnKooij:
Any advise for me? I was trying to get the coverage back to 100%, R# dotCover says I covered everything but I suspect coveralls is a bit more strict? If I had to guess than I would need to add a test where request is null in ClientCertificateWebClient.GetWebRequest and that seems both quite difficult as well as not really useful?

In my opinion testing the WebClient that will add certificates is a bit over the top to begin with. But if you're OK with the testable class I had to create then we are indeed covered better now.

AndersAbel:
It's possible to click through from the notification to the report from coveralls - which is based on Visual Studio's code coverage. https://coveralls.io/builds/11561088/source?filename=%2FKentor.AuthServices%2FInternal%2FClientCertificateWebClient.cs

You're right in your guess that it's the ? in GetWebRequest that is the problem.

If there is any situation where GetWebRequest could return null, then a specific null check should be added. If that's never going to happen, just drop the ?.

MartijnKooij:
No, as far as I know there is no situation where the web request would be null. I just blindly followed R#'s advice and added the conditional access.
I cannot find any official documentation on this however, the MSDN example also just accesses the request's properties without checking... I think we're good.

desjoerd:
I've fixed the merge conflicts by consolidating the dependencies of the sample applications. They are now all on the same owin versions. I hope this can be merged into Authservices now ^^

MartijnKooij:
Pfff... Could not get the merge clean so I just reimplemented the change in the current master.
This should now again be a clean PR with just the changes to add support for a TLS on the artifact resolve. Could you please merge this so I (and by now 2 other Dutch companies apparently) can get this item out of their own custom forks?

jbreuer:
Please merge this PR into the master. Our company also connects to DigiD and I've run into the same issue. Now running on a fork of https://github.com/PGGMTeam5/authservices.

MartijnKooij:
I have just updated my merge once more so all is clean once more.
I have also transferred my repository (ad thus this merge) to the account of the team at my current job. I will be leaving the company at the end of this month and I really really hope that I can close this issue before then. If not for the company than at least for my peace of mind :)
By now I have counted 6 Dutch companies that require this change in order to connect the Dutch IdP. Is there anything I can do to help get this PR in?

Kind regards,
Martijn Kooij

jbreuer:
Thanks for fixing the merge issue! One more small suggestion. 

Please give the option to also set on ArtifactResolutionTlsCertificate: https://github.com/PGGMTeam5/authservices/blob/master/Sustainsys.Saml2/Configuration/SPOptions.cs#L290

`public X509Certificate2 ArtifactResolutionTlsCertificate { get; set; }`

I don't read the certificate from a config file, but I set it in my startup.cs file. I've added this change to my own fork.

MartijnKooij:
I don't see any harm in adding the setter, sounds like a valid way to use it.

jbreuer:
Thanks for adding the setter. It seems the AppVeyor build failed. Not sure why.

jbreuer:
This has been fixed in the master branch and will soon be released as a NuGet package. 

Instead of setting the certificate as ArtifactResolutionTlsCertificate on the options it's now possible to add it like this:

`optionsIdpDigid.SPOptions.ServiceCertificates.Add(new ServiceCertificate { Use = CertificateUse.TlsClient, Certificate = certDigid });`

So you add a service certificate with the TlsClient enum.

jbreuer:
Version 2.2 is out https://www.nuget.org/packages/Sustainsys.Saml2/ :-)

PR #753, opened 2017-06-27
Added possibility to set Certificate Password in config

MaurizioPz:


explunit:
I'm not the project owner but it's my opinion that this feature does not need to be in the core library and that adding it might encourage weak certificate management practices. The options classes accept `X509Certificate2` object so if you have a different process (i.e. not using windows cert store) for obtaining the certificates you can do that outside the library.

PR #784, opened 2017-08-31
Allows for cookie lifetime to be set by persistentSessionLifetime

### LINKED ISSUE #242: Persist session across browser sessions
Is there a way to persist the authentication cookie ('FedAuth') between browser sessions? My goal is to keep the user logged in after he restarts his browser. I'm using the MVC Module. 

After adding `sessionToken.IsPersistent = true;` on line 72 in [this file](https://github.com/KentorIT/authservices/blob/069841535fc33655984d1cf1cdad63802eb856a4/Kentor.AuthServices.Mvc/CommandResultExtensions.cs) the cookie persists but that requires me to modify your code. 

Maybe I'm missing something and it's also possible in other ways?

jsoncox:
Resolves #242. Allows for cookie lifetime to be set by persistentSessionLifetime of cookieHandler in federationConfiguration.

### DIFF
--- a/Kentor.AuthServices.HttpModule/CommandResultHttpExtension.cs
+++ b/Kentor.AuthServices.HttpModule/CommandResultHttpExtension.cs
@@ -124,7 +124,10 @@ namespace Kentor.AuthServices.HttpModule
                     null,
                     DateTime.UtcNow,
                     commandResult.SessionNotOnOrAfter ??
-                    CalculateSessionNotOnOrAfter());
+                    CalculateSessionNotOnOrAfter())
+                {
+                    IsPersistent = true
+                };
 
                 EnsureSessionAuthenticationModuleAvailable();
 


explunit:
Thanks for looking into this. However, it would need to be a configuration setting since not everyone will want the cookies to be persistent.

Also, you need to add tests to cover any new code that you add.

jamestelfer:
I may not fully understand the existing code, but it appears that the current code has the following precedence for setting the expiry date:
* the IdP assertion content
* the .NET configuration
* the .NET default

In the first two cases (IMO), the default behaviour should be to set the cookie to persistent, as otherwise the expiry will not be honoured. It might be good to be able to opt out of this behaviour, but at the moment any expiry is ignored as the cookie is never persistent.

In the third case, the default could be to use a session cookie, as no lifetime has been specified.

Complicating this persistent/session cookie discussion is the fact that in most current browsers under common configuration settings, session cookies are not discarded when the tab/browser is closed and restarted. This has the effect of making the cookies persist forever.

The thing that I'm not sure about is whether there is a claim internal to the persisted `ClaimsIdentity` that will expire the cookie after a particular date/time. This would invalidate what I said in the previous paragraph.

jamestelfer:
Thanks @AndersAbel, I had a feeling that my knowledge was lacking there. Many browsers don't delete session cookies any more on restart, depending on the browser settings -- if the user opts to keep their tabs, for example.

However, if the session cookie has its own timebomb inside then this issue is mitigated.

AndersAbel:
@jamestelfer Yes, it is somewhat mitigated. But not completely. The server won't accept the session cookie after it's expired, but the cookie will hang on in the browser. Which could cause issues with the cookie size. I'll look closer when I get to reviewing this PR. First I have to get CI up and running again after the repo rename/transfer.

Issue #812, opened 2017-10-12
Hard coded encryption algorithms prevent using this package in a FIPS environment


mcnallys:
There are 2 different areas of the code preventing correct operation on a FIPS compliant machine.

"this implementation is not part of the Windows Platform FIPS validated cryptographic algorithms."

This stems from using the  RijndaelManaged symetricAlgorithm in CryptographyExtensions and some other issues that can be corrected in RSAEncryptedXML  using some of the code from

http://referencesource.microsoft.com/#System.Configuration/System/Configuration/FipsAwareEncryptedXml.cs,fad55741da0e4b5f

Would you be willing to accept a pull request to fix these issues?
When do you expect to release an update to nuget?

AndersAbel:
Yes, I'd be willing to accept a PR to fix this. And I appreciate to have a discussion before you go ahead. One potential problematic place is the fix for when the cryptographic service provider associated with the certificate doesn't support SHA256: https://github.com/Sustainsys/Saml2/blob/c66a835af1a4acfd46051cf95a19a4093e15bf2f/Kentor.AuthServices/Internal/CryptographyExtensions.cs#L63 

For a Nuget release I'd say ASAP, but I've too much to do right now so can't promise a date.

If you want to get prioritized PR review and release coordination, that's something I can offer as part of my commercial services.

agrady42:
I'm interested in this as well - 

For discussion - is there a way to set the algorithm used with configuration (even if it's something as simple as "useFipsCompliantAlgorithm=true" and document that this will cause problems with providers that don't support SHA256)?

For the rest - is there anything I can do to help with the PR?

agrady42:
Open mouth, insert foot. Somehow I missed PR#813 just above my comment - thanks @mcnallys !

Just waiting on the release then.

AndersAbel:
Fixed in v2.

PR #813, opened 2017-10-13
Fips compliance

mcnallys:
Enables the configuration of the encryption algorithm , which can be used to switch to fips compliant algorithms.

This branch was created from the v0.21.2  tag.

#812

### DIFF
--- a/Kentor.AuthServices/Internal/CryptographyExtensions.cs
+++ b/Kentor.AuthServices/Internal/CryptographyExtensions.cs
@@ -28,7 +28,7 @@ namespace Kentor.AuthServices.Internal
 
             var encryptedXml = new EncryptedXml();
             byte[] encryptedElement;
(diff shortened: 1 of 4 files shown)


AndersAbel:
I see you used the `[ExcludeFromCodeCoverage]` attribute. Is there no way to test this? Or break the method apart and only exclude as few parts as possible from tests?

I understand that this kind of code is hard to test as it depends on machine state (and I definitely want the tests to pass on a clean machine without having to do special config). But if there's a way to do at least some tests that would be good.

mcnallys:
The main reason I went that way is like you said due to being based on machine state and the fact that the existing code in that class was already excluded.

ckem-exigerfed:
Hello, am interested in implementing this change to allow for FIPS compliance. Is there a plan to merge in this PR?

PR #822, opened 2017-10-21
Use SameSiteMode.None for cookies in ASP.NET Core handler

### LINKED ISSUE #809: Relaystate cookie lost in ASP.NET Core handler
Hello,
Using Sustainsys.Saml2.AspNetCore2.

When /Saml2 gets posted from Okta HandleRequestAsync responds to browser with a file.

It's my first time dealing with this protocol., i always used OpenId. 

How should I debug this?

commandName = context.Request.Path.Value.Substring(
                    options.SPOptions.ModulePath.Length).TrimStart('/');

returns "", which means give me the file., but why? is Okta giving the wrong command? (blank) or this file should be handled somehow?

My code is very simple and does not have anything fancy. Tried specifying ReturnUrl but does nothing.

Cheers

explunit:
Fixes #809 

The default became SameSiteMode.Lax in ASP.NET Core 2, which means that only GET requests from other site will be allowed to send the cookies.

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/CommandResultExtensions.cs
+++ b/Sustainsys.Saml2.AspNetCore2/CommandResultExtensions.cs
@@ -30,7 +30,13 @@ namespace Sustainsys.Saml2.AspNetCore2
                 httpContext.Response.Cookies.Append(
                     commandResult.SetCookieName,
                     cookieData,
-                    new CookieOptions() { HttpOnly = true } );
+                    new CookieOptions()
+                    {
+                        HttpOnly = true,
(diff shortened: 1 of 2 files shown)


andreujuanc:
Makes sense. Such small change, but big impact.

I was not aware about this default in aspnetcore2. 

Can't test with my project right now, but will do so in few days.

andreujuanc:
I Just test it and works! 
Thanks

AndersAbel:
Thanks for picking that up @explunit, I thought it was a really weird behaviour.

PR #824, opened 2017-10-24
Additional settings for a custom sp

bananaguardyourself:


AndersAbel:
To be able to review this, it would help if you could add an overall description of the new features/changes in the PR directly. That way it will be easier for me when I read the code.

Also, there's a code coverage indication that the tests are not covering all lines. If you click the "details" link on the coverage/coveralls line, you can click through to see what is not used by the unit tests.

AndersAbel:
Thanks for the update so that it passes the build checks. Now I only need a description on a high level what the changes are good for.

dzinin:
We have a project with a buggy SP which we can't fix for various reasons. And since **stubidp** is a _stub_ we decided to alter SAML responses in a way that our customer's SP can consume them.

These are two issues which we are trying to solve with this PR:
1. Only POST binding should be returned in _metadata_ endpoint
2. Attributes must have _xsi:type="xs:string"_

And as you can see we've also added two _boolean_ fields to the JSON schema which are _false_ by default. This guarantees that these changes are backwards compatible and won't affect the others.

Kentor.AuthServices.StubIdp/Models/MetadataModel.cs
```diff
@@ -10,7 +10,7 @@ namespace Kentor.AuthServices.StubIdp.Models
 {
     public static class MetadataModel
     {
-        public static ExtendedEntityDescriptor CreateIdpMetadata(bool includeCacheDuration = true)
+        public static ExtendedEntityDescriptor CreateIdpMetadata(bool defaultPost = false, bool includeCacheDuration = true)
```
AndersAbel: Two bool params with default values look a bit error prone, especially when you add the new one first. Please remove the defaults (for both of them)

Kentor.AuthServices.StubIdp/Models/MetadataModel.cs
```diff
@@ -63,7 +65,7 @@ public static ExtendedEntityDescriptor CreateIdpMetadata(bool includeCacheDurati
             return metadata;
         }
 
-        public static ExtendedEntitiesDescriptor CreateFederationMetadata()
+        public static ExtendedEntitiesDescriptor CreateFederationMetadata(bool defaultPost = false)
```
AndersAbel: Same as above, remove the default value.

Kentor.AuthServices/SAML2P/Saml2Response.cs
```diff
@@ -208,6 +209,7 @@ private void ReadAndValidateInResponseTo(XmlElement xml, Saml2Id expectedInRespo
             Saml2Id inResponseTo,
             string relayState,
             Uri audience,
+            bool enforceXmlns,
```
AndersAbel: This is part of the main library's public API, so this is a breaking change. You have to add another overload that keeps the existing signature to not break existing code.

Kentor.AuthServices/Saml2StatementExtension.cs
```diff
@@ -69,9 +80,18 @@ private static XElement ToXElement(Saml2AttributeStatement attributeStatement)
...
                 attributeElement.AddAttributeIfNotNullOrEmpty("OriginalIssuer", attribute.OriginalIssuer);
 
+                XNamespace xs = @"http://www.w3.org/2001/XMLSchema";
+                XNamespace xsi = @"http://www.w3.org/2001/XMLSchema-instance";
+
                 foreach (var value in attribute.Values)
                 {
-                    attributeElement.Add(new XElement(Saml2Namespaces.Saml2 + "AttributeValue", value));
+                    if (enforceXmlns)
+                        attributeElement.Add(new XElement(Saml2Namespaces.Saml2 + "AttributeValue", value,
```
AndersAbel: Please use {} on all if statements. Also rewrite this to create the new XElement first and then add the XNamespace attributes inside an if-block. Don't want the basic attributeElement.Add line duplicated as it can make bugs appear if one is changed and not the other. And having it twice means tests must be duplicated to ensure functional coverage.

Kentor.AuthServices.StubIdp/App_Data/default.json
```diff
@@ -1,5 +1,7 @@
 {
   "HideDetails": false,
+  "EnforceXmlns": false,
```
AndersAbel: This enforces XS (Xml Schema), not NS (NameSpace). And it is for attributevalues only, which should be more clear. What about `EnforceAttributeXsString`?

Tests/Tests/Saml2AssertionExtensionsTests.cs
```diff
@@ -183,5 +183,61 @@ public void Saml2AssertionExtensions_ToXElement_Statements()
 
             actualAuthnXml.Should().BeEquivalentTo(expectedAuthnXml);
         }
+
+        [TestMethod]
+        public void Saml2AssertionExtensions_ToXElement_Statements_Xmlns()
```
AndersAbel: Clean up this test and make sure it only test the XS declaration. The authnInstant checks are not relevant for this test. I guess that's something that's left from a test you copied?

Tests/Tests/Saml2P/Saml2ResponseTests.cs
```diff
@@ -1778,6 +1779,7 @@ public void Saml2Response_Xml_FromData_ContainsAudienceRestriction()
                 new Saml2Id("InResponseToID"),
                 null,
                 new Uri(audience),
+                false,
```
AndersAbel: Revert this change. Having to change this is a sign of changing the public API (as commented previously)

Tests/Tests/Saml2StatementExtensionsTests.cs
```diff
@@ -31,6 +31,17 @@ public void Saml2StatementExtensions_ToXElement_TypeCheck()
                 .And.Message.Should().Be("Statement of type StubSaml2Statement is not supported.");
         }
 
+        [TestMethod]
```
AndersAbel: What does this test that isn't already covered by other tests?

bananaguardyourself: According to coveralls.io, code is not fully covered without this test:

https://coveralls.io/builds/13864802/source?filename=%2FKentor.AuthServices%2FSaml2StatementExtension.cs

bananaguardyourself:
Thank you for your response! There is a new commit with all your comments taken into account.

AndersAbel:
Force pushed an update to not contain a merge of master. The PR should be kept clean. The right way to refresh it with regards to master is to do a rebase. But in this case I'll handle it when merging to master.

AndersAbel:
Started on reviewing this, but there's still issues left (I accidentally pushed approve to all review comments, which is not correct).

Talked directly to customer rep and this has been put on hold for now.

PR #846, opened 2017-11-15
Fix broken links to web.config samples

neugenes:
I just noticed these links were broken, so I modified them to point to what I assume are the correct files.


explunit:
thanks!

PR #919, opened 2018-02-27
Incorrect IdentityServer4 reference URL

robertpmansion:


AndersAbel:
Thanks for noticing and fixing!

PR #590, opened 2016-11-14
Feature/allow unsigned logout request/ response

MinhNguyenDev:
This is the solution for issue #446.

Background story:
I tried to integrate OneLogin Single Sign On to our application:
(https://support.onelogin.com/hc/en-us/articles/202673944-How-to-Use-the-OneLogin-SAML-Test-Connector)
The issue is when OneLogin sends a saml logout request or response, the saml message doesn't contain the signature and Kentor threw an exception because of that.
This change introduces 2 optional settings to skip validating the log out request/response to work around the fact that some IDP doesn't support signed log out saml message:
allowUnsignedLogOutRequest
allowUnsignedLogOutResponse

Example :

    <identityProviders>
      <add entityId="https://app.onelogin.com/saml/metadata"
           signOnUrl="one login sso log in url"
           logoutUrl="one login sso log out url"
           allowUnsignedLogOutRequest="true" 
           allowUnsignedLogOutResponse="true" 
           binding="HttpRedirect">
            ....
      </add>
    </identityProviders>

Issue #446, opened 2016-04-07
Wrong issuer in dynamically created sp single logout


lindea:
Hi!

I have a problem with the single signout functionality.
I'm using  AS.Owin  1.7.2 on a site with multiple domains (actually  it's several sites in same application).
I create SP for all domains dynamically like this:

```
foreach(var domain in domains){
 app.UseKentorAuthServicesAuthentication(CreateAuthServicesOptions("kentor", domain.Name));
}
```
This works fine for login, and the login request is called with the right metadata based on the domain.
For the logout request the issuer is always from the first domain, and thus fails. 
Each SP has it's own unique module path.

For the logout I use
`OwinContext.Authentication.SignOut(DefaultAuthenticationTypes.ApplicationCookie, DefaultAuthenticationTypes.ExternalCookie);`

This doesn't seem to trigger single sign out, so I redirect to <ModulePath>/logout where the signout request is triggered.

Any suggestions?

AndersAbel:
You need to include AuthServices in the sign out types. Either sign out of everything by calling `OwinContext.Authentication.SignOut()` or include the signin-type of the AuthServices instance that you want to sign out of.

lindea:
Thank you for The super fast response!
I was able to trigger signout by specifying the AuthenticationType as you suggested. The problem now is that the respons is required to be signed, which it's not. Is there an option to turn off that requirement?

AndersAbel:
There's no option to turn of the signing requirement for logout messages. It's part of the [profile spec](https://www.oasis-open.org/committees/download.php/56782/sstc-saml-profiles-errata-2.0-wd-07.pdf) section 4.4.4.2:

> The responder MUST authenticate itself to the requester and ensure message integrity, either by signing the message or using a binding-specific mechanism.

But if you really need an option to switch it off, I can consider adding it.

lindea:
Thanks again. I'll have a talk with the idp-guys first and hear what they have to say.

lindea:
Hi again!
I've talked to the Idp, and they do not provide signed logout response despite the requirements. They might do in the future though. It would be nice to be able to choose whether or not to accept unsigned logon responses, so I'll be grateful if you would consider implementing this.

AndersAbel:
I see the need for it, but as this is very non-standard it's far down on the priority list for free-time development. If you want to sponsor the development to get it done faster, please drop me a mail.

PR #924, opened 2018-02-28
Add new identity provider configuration to allow unsigned log out request/response

cederlof:
Code from @MinhNguyenDev PR: https://github.com/Sustainsys/Saml2/pull/590 which is based on an old version of Saml2 (and had conflicts with master).

This is the solution for issue #446, adding two new parameters to the IdentityProvider-element in config:
allowUnsignedLogOutRequest and allowUnsignedLogOutResponse. For use when the IDP does not support signing of log out request/response.

```

<identityProviders>
  <add entityId="https://app.onelogin.com/saml/metadata"
       signOnUrl="one login sso log in url"
       logoutUrl="one login sso log out url"
       allowUnsignedLogOutRequest="true" 
       allowUnsignedLogOutResponse="true" 
       binding="HttpRedirect">
        ....
  </add>
</identityProviders>
```

### DIFF
--- a/Sustainsys.Saml2/Configuration/IdentityProviderElement.cs
+++ b/Sustainsys.Saml2/Configuration/IdentityProviderElement.cs
@@ -137,6 +137,36 @@ namespace Sustainsys.Saml2.Configuration
         }
 
         /// <summary>
+        /// Allow log out response without signature
+        /// If true log out response without signature is still accepted
+        /// If false log out response without signature is not accepted
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1726:UsePreferredTerms", MessageId = "LogOut")]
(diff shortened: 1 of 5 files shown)


ricardosaracino:
Question, is this going to be rolled in ever?

cederlof:
Hi! The fix resolves the feature for unsigned logout responses, however not allowing unsigned logout requests.

AndersAbel:
@cederlof Yes, that's true. And per the discussion in the other PR we limited it to responses for security reasons. Allowing unsigned logout requests open up for some kinds of attacks and I am very reluctant to add compatibility flags that allows unsecure usage of the library.
What is the background for the requirement of unsigned logout requests?

cederlof:
@AndersAbel I agree with your reasoning! We do not have a requirement for unsigned logout requests, but I thought it would be good to mention as this PR was labeled `duplicate`.

PR #1016, opened 2018-09-27
`CertificateUse.Both` missing word in summary tag

duncansmart:


AndersAbel:
Thanks, nice catch. I noticed I failed to get the xml doc included in the nugets, will make an update with that (guess  I'll have a few more bugs coming in first though)

PR #1020, opened 2018-10-05
add support for idp selection

sandorfr:
Add a way to select Idp to be used in a similar fashion as https://github.com/Sustainsys/Saml2/blob/master/docs/owin-middleware.rst#selecting-idp

```csharp
var properties = _signInManager.ConfigureExternalAuthenticationProperties(samlScheme, redirectUrl);
properties.Items.Add("idp", method.GetConfiguration<Saml2IdpConfiguration>().EntityId);
return new ChallengeResult(samlScheme, properties);
```

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/Saml2Handler.cs
+++ b/Sustainsys.Saml2.AspNetCore2/Saml2Handler.cs
@@ -67,10 +67,10 @@ namespace Sustainsys.Saml2.AspNetCore2
 
         private string CurrentUri
         {
-            get => context.Request.Scheme + "://" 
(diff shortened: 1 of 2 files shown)


AndersAbel:
Good catch, that should of course be included.

There are some extra changes that shouldn't be in the PR, please see the diff. Before merging I want to have a test for the feature was well. There is one for owin, so it can probably be used as a base.

sandorfr:
I'll do those things on Monday, thanks for the quick answer.

sandorfr:
Hi @AndersAbel, I've removed the unwanted changes on VersionInfo and added a test.

PR #1041, opened 2018-11-14
Introduce the Strategy pattern to IdentityProviderDictionary

jliu-appliedsystems:
Introduce the Strategy pattern to IdentityproviderDictionary in order to allow clients of this package to bring their own implementations. This provides a solution to issue #1042.


AndersAbel:
I'm sorry, but this is not a change I will accept at this time. The reason is that I'm not happy with the current design, and planning for a larger rewrite for 3.0 where this will be addressed.

For now, there are notifications that you can implement that lets you return an Idp object instead of accessing the IdentityProviderDictionary (at least in most places, for some bindings/ edge cases it does not work with this design)

PR #1048, opened 2018-12-24
Add support to use the RelayState as the return url.

diegonzls:
Use the RelayState as the ReturnUrl when AllowUnsolicitedAuthnResponse is enabled. To keep compatibility and not break any current behaviour a new setting has been added to indicate that a ReturnUrl from the IDP is expected in the RelayState. This should be a fix for https://github.com/Sustainsys/Saml2/issues/250

### DIFF
--- a/Sustainsys.Saml2/IdentityProvider.cs
+++ b/Sustainsys.Saml2/IdentityProvider.cs
@@ -83,7 +83,7 @@ namespace Sustainsys.Saml2
                 Validate();
             }
 
-            RelayStateUsedAsReturnUrl = config.RelayStateUsedAsRedirectUrl;
+            RelayStateUsedAsReturnUrl = config.RelayStateUsedAsReturnUrl;
         }
 
         private void Validate()
@@ -257,7 +257,7 @@ namespace Sustainsys.Saml2
         public bool AllowUnsolicitedAuthnResponse { get; set; }
 
         /// <summary>
-        /// Does the RelayState contains the redirect url?, 
+        /// Does the RelayState contains the return url?, 
         /// This setting is used only when the AllowUnsolicitedAuthnResponse setting is enabled.
         /// </summary>
         public bool RelayStateUsedAsReturnUrl { get; set; }
--- a/Sustainsys.Saml2/IdentityProvider.cs
+++ b/Sustainsys.Saml2/IdentityProvider.cs
@@ -82,6 +82,8 @@ namespace Sustainsys.Saml2
             {
                 Validate();
             }
+
+            RelayStateUsedAsReturnUrl = config.RelayStateUsedAsRedirectUrl;
         }
 
         private void Validate()
@@ -253,6 +255,12 @@ namespace Sustainsys.Saml2
         /// Is this idp allowed to send unsolicited responses, i.e. idp initiated sign in?
         /// </summary>
         public bool AllowUnsolicitedAuthnResponse { get; set; }
+
+        /// <summary>
+        /// Does the RelayState contains the redirect url?, 
+        /// This setting is used only when the AllowUnsolicitedAuthnResponse setting is enabled.
+        /// </summary>
+        public bool RelayStateUsedAsReturnUrl { get; set; }
 
         private string metadataLocation;
 
--- a/Sustainsys.Saml2/WebSSO/AcsCommand.cs
+++ b/Sustainsys.Saml2/WebSSO/AcsCommand.cs
@@ -55,7 +55,7 @@ namespace Sustainsys.Saml2.WebSso
 
                     var result = ProcessResponse(options, samlResponse, request.StoredRequestState, idpContext, unbindResult.RelayState);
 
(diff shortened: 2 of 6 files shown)


Sustainsys.Saml2/WebSSO/AcsCommand.cs
```diff
@@ -88,6 +95,17 @@ public CommandResult Run(HttpRequestData request, IOptions options)
             throw new NoSamlResponseFoundException();
         }
 
+        private static IdentityProvider GetIdpContext(XmlElement xml, IOptions options)
+        {
+            var entityId = new EntityId(xml["Issuer", Saml2Namespaces.Saml2Name].GetTrimmedTextIfNotNull());
+
+            IdentityProvider identityProvider;
+
+            options.IdentityProviders.TryGetValue(entityId, out identityProvider);
```
AndersAbel: Use the notification instead, to allow consumers to overload Idp lookup (which allows to bypass the IdentityProviders collection completely).

Sustainsys.Saml2/WebSSO/AcsCommand.cs
```diff
@@ -45,16 +46,22 @@ public CommandResult Run(HttpRequestData request, IOptions options)
...
 
                     var samlResponse = new Saml2Response(unbindResult.Data, request.StoredRequestState?.MessageId, options);
+                    
+                    var idpContext = GetIdpContext(unbindResult.Data, options);
+
+                    var result = ProcessResponse(options, samlResponse, request.StoredRequestState, idpContext, unbindResult.RelayState);
 
-                    var result = ProcessResponse(options, samlResponse, request.StoredRequestState);
-                    if(unbindResult.RelayState != null)
+                    if(unbindResult.RelayState != null && (idpContext != null ? !idpContext.RelayStateUsedAsReturnUrl : false))
```
AndersAbel: This expression doesn't look right. What if SP-initiated sign in is used with an Idp that has RelayStateUsedAsReturnUrl=true? An Idp could support both SP-initiated sign in (where RelayState should be just relayed) and Idp-initiated where the RelayState should be treated as a ReturnUrl.

Sustainsys.Saml2/IdentityProvider.cs
```diff
@@ -82,6 +82,8 @@ internal IdentityProvider(IdentityProviderElement config, SPOptions spOptions)
             {
                 Validate();
             }
+
+            RelayStateUsedAsReturnUrl = config.RelayStateUsedAsRedirectUrl;
```
AndersAbel: Please use one term consistently - either "redirect" or "return"

Sustainsys.Saml2/WebSSO/AcsCommand.cs
```diff
@@ -96,11 +114,15 @@ public CommandResult Run(HttpRequestData request, IOptions options)
...
             Saml2Response samlResponse,
-            StoredRequestState storedRequestState)
+            StoredRequestState storedRequestState,
+            IdentityProvider identityProvider,
+            string relayState)
         {
             var principal = new ClaimsPrincipal(samlResponse.GetClaims(options, storedRequestState?.RelayData));
 
-            if(options.SPOptions.ReturnUrl == null)
+            var isRelayStateUsedAsReturnUrl = identityProvider != null ? identityProvider.RelayStateUsedAsReturnUrl : false;
```
AndersAbel: See comment above about an Idp that is configured to use both SP-initiated (Relaystate is relaystate) and Idp-initiated with (Relaystate is returnurl).

Sustainsys.Saml2/WebSSO/AcsCommand.cs
diegonzls: I have updated the code to use Notification instead.

Sustainsys.Saml2/WebSSO/AcsCommand.cs
diegonzls: By using Saml2Notifications.GetIdentityProvider the idpContext is not going to be null and so the expression can be simplified. Please see the latest commit.

Sustainsys.Saml2/IdentityProvider.cs
diegonzls: Amended. I am now using "return" everywhere.

Sustainsys.Saml2/WebSSO/AcsCommand.cs
diegonzls: See comments above.

diegonzls:
@AndersAbel  Could you please re run the "Checks" on my last commit 6a155ef 

Running the unit test locally is fine and I don't see how the change on that commit would affect the test that is failing. Thanks.

AndersAbel:
Thanks for updating from my first review. I've looked through this and found an open redirect issue. So I've added URL validation of absolute return urls as well as allowing relative URLs.

Can you please have a look if you think this makes sense? If it does, I think this is ready to merge.

diegonzls:
I have checked the change that you did and I think that it is ok. Thank you for your help.

Issue #1049, opened 2019-01-08
IDPSSODescriptor element is not deserialized when after RoleDescriptor element of type ApplicationServiceType or SecurityTokenServiceType


veselkamartin:
Using latest version of source codes.
Expected: IDPSSODescriptor is deserialized
Actual: IDPSSODescriptor is skiped during deserialization and ignored

There is bug in metadata deserialization.
When there is RoleDescriptor element of type ApplicationServiceType or SecurityTokenServiceType before IDPSSODescriptor element, the IDPSSODescriptor is not parsed and is not deserialized.
Example of simplified metadata xml, which is wrongly deserialized without IDPSSODescriptor:
```
  <EntityDescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" ID="_4e110583-2a5e-4100-8bff-6189f61975be" entityID="urn:microsoft:cgg2010:fpsts">
    <Signature xmlns="http://www.w3.org/2000/09/xmldsig#">
      ...
    </Signature>
    <RoleDescriptor xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:fed="http://docs.oasis-open.org/wsfed/federation/200706" xsi:type="fed:SecurityTokenServiceType" protocolSupportEnumeration="http://docs.oasis-open.org/wsfed/federation/200706">
      ...
    </RoleDescriptor>
    <IDPSSODescriptor protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol">
      ...
    </IDPSSODescriptor>
  </EntityDescriptor>
```
Full metadata xml causing problem:
https://tnia.eidentita.cz/fpsts/FederationMetadata/2007-06/FederationMetadata.xml

The bug is in file Metadata\MetadataSerializer.cs in method "protected virtual EntityDescriptor ReadEntityDescriptor(XmlReader reader, SecurityTokenResolver tokenResolver)" on lines 2015 and 2020.

PR #1050, opened 2019-01-08
Fixing issue #1049 - IDPSSODescriptor element is not deserialized

veselkamartin:
…n after RoleDescriptor element of type ApplicationServiceType or SecurityTokenServiceType


AndersAbel:
Would it be possible to turn the example metadata you outlined in #1049 and create a test based on that and include that?

veselkamartin:
I have added simple unit test for specified problem.

AndersAbel:
Thanks for the test. I run it without applying the patch and now it is obvious what the problem is.

Thanks a lot for taking the time to find and fix the bug!

PR #1056, opened 2019-01-18
Make PostConfigureSaml2Options publicly accessible

GeertHuls:
Hello,

I would like to have the PostConfigureSaml2Options to be publicly accessible so that it can be registered and used in other projects. I've tested this solution and everything works as expected.

Thanks in advance.

Regards.

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/PostConfigureSaml2Options.cs
+++ b/Sustainsys.Saml2.AspNetCore2/PostConfigureSaml2Options.cs
@@ -1,16 +1,11 @@
-﻿using Sustainsys.Saml2;
+﻿using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System;
-using System.Collections.Generic;
-using System.Linq;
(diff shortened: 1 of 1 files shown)


AndersAbel:
I'm a bit curious why/how you would need to use PostConfigureSaml2Options from other projects? Can you elaborate a bit on the use case?

In general I try to be restrictive about making things `public` because anything that is `public` needs to be kept compatible within a major version. So from a API stability code flexibility perspective, the less things that are `public` the better.

GeertHuls:
Oh I see. Well I figured using the PostConfigureSaml2Options in order to change the saml2 scheme dynamically instead of settings them at startup only.
That way, we can allow the users of our application to switch to a different adfs server themselves if they want to.

A solution to this problem is explained here, using the IAuthenticationSchemeProvider amongst others:
https://stackoverflow.com/questions/48910070/adding-new-authentication-schemes-dynamically

But on second thought, creating my own PostConfigureSaml2Options does also work apparently so that means I won't need the PostConfigureSaml2Options from this library anymore. In that case you may discard this pr!

Regards.

PR #1060, opened 2019-01-27
Add TokenValidationParametersCreated notification

explunit:
This allows customization of audience, issuer, etc. beyond what the core library would support.

### DIFF
--- a/Sustainsys.Saml2/Configuration/Saml2Notifications.cs
+++ b/Sustainsys.Saml2/Configuration/Saml2Notifications.cs
@@ -1,4 +1,5 @@
-﻿using Sustainsys.Saml2.Metadata;
+﻿using Microsoft.IdentityModel.Tokens;
+using Sustainsys.Saml2.Metadata;
 using Sustainsys.Saml2.Saml2P;
 using Sustainsys.Saml2.WebSso;
 using System;
@@ -142,5 +143,12 @@ namespace Sustainsys.Saml2.Configuration
         /// </summary>
         public Func<EntityId, IDictionary<string, string>, IOptions, IdentityProvider> GetIdentityProvider { get; set; }
             = (ei, rd, opt) => opt.IdentityProviders[ei];
+
+        /// <summary>
+        /// Notification called when the token handler has populated the
+        /// <see cref="TokenValidationParameters"/>. Modify it's properties to customize
+        /// the generated validation parameters.
+        /// </summary>
+        public Action<TokenValidationParameters> TokenValidationParametersCreated { get; set; } = tvp => { };
     }
 }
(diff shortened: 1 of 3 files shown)


AndersAbel:
Yes, the `TokenValidationParameters` need to be exposed for the end user to alter. The way the other protocol handlers this is however done in another way: They have a `TokenValidationParameters` instance as part of the options. For each validation operation that instance is then cloned to a new instance. The per-request parameters are then set on that cloned instance. E.g. the `RequireSignedTokens` parameter (if we use the built in signature validation) would be set to false for a token received through a signed message.

That doesn't exclude the possibility to have a notification like this though - a notification would give the user a last chance to alter the settings before they are used. I'm not sure that is a good idea though. Because this would open for the kind of "bug fixes" where someone would fix a signature problem by setting `RequiredSignedTokens` to false.

What do you think? Would a "template" `TokenValidationParameters` in the options solve most use cases?

explunit:
It would solve most, but I think there are some cases that it would not. For example, the current code directly sets a single expected audience but if someone wanted to allow multiple audience values they would have no ability to do that.

The above might be a contrived example (and contrary to the SAML spec), but looking at the current code that's the sequence of calls that I would worry about.

AndersAbel:
I think we should keep the idea of implementing a "template" `TokenValidationParameters` in the future, keeping in line with the other protocol handlers.

But I still think this PR has value. However I'm considering if it should somehow be named in a way to indicate that using it makes it possible to do unsafe things.

One idea would be to put it under a a separate sub-property in the `Saml2Notifications` class: `Saml2Notifications.Unsafe.TokenValidationParametersCreated`. We could also do a similar thing in the compatibility section.

Further I think that we should consider including some kind of state in the param list. I've found out that it's a general flaw with the notifications. In many cases there's no way to get any additional state or info about the current request/SAML message/Active Idp which makes it hard to implement conditional checks/modifications.

explunit:
I've made more state available to the notification (IDP, XmlElement). I was undecided about `XmlElement` but I see that it's already exposed as a public property. Let me know if you disagree about sending this to the notification.

Regarding naming: I'm not opposed to moving it, but I'm not sure about "unsafe". I don't see how modifying the `TokenValidationParameters` at this stage is less safe than providing a template that sets `ValidateAudience` to false (for example).

AndersAbel:
Even with the template, I think that the most important validation params will be validated/set. E.g. the signature validation params will be set for each request (I do plan to move to use the built in signature validator instead of `SignedXml`). So even with a template, I would try to add some kind of safety net against the most stupid configurations. The notification would then provide a "do whatever you want at your own risk" possibility. That's why I'm considering a name indicating that it is dangerous.

And once more this is a discussion about a potential future that might or might not become reality. The problem is that once we release this as a public API we're stuck with the naming.

explunit:
OK, makes sense. I've moved it to a new `Notifications.Unsafe` section.

Issue #780, opened 2017-08-21
Compatibility setting for authentication context parsing


explunit:
An issue I've seen frequently is a parsing error when the IDP sends something like this:

```
      <saml:AuthnContext>
        <saml:AuthnContextClassRef>phonefactorPREAUTH/contract</saml:AuthnContextClassRef>
        <saml:AuthnContextDeclRef>phonefactorPREAUTH/contract</saml:AuthnContextDeclRef>
      </saml:AuthnContext>
```
We get error like this:
```
Kentor.AuthServices.Exceptions.BadFormatSamlResponseException: The SAML response contains incorrect XML ---> System.Xml.XmlException: ID0013: The value must be an absolute URI.
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadSimpleUriElement(XmlReader reader, UriKind kind, Boolean allowLaxReading)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadAuthenticationContext(XmlReader reader)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadAuthenticationStatement(XmlReader reader)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadAssertion(XmlReader reader)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadToken(XmlReader reader)
   at Kentor.AuthServices.Saml2P.Saml2Response.<CreateClaims>d__57.MoveNext()
```

IDP side is often reluctant to fix this, as they consider their response to be conforming with the spec, e.g. see here: https://www.novell.com/support/kb/doc.php?id=7015106

Would it make sense to add a `Compatibility.IgnoreAuthenticationContextInResponse` setting to simply skip parsing of this section?

I believe it would be as simple as overriding `ReadAuthenticationContext` in `Saml2PSecurityTokenHandler` and returning an empty `Saml2AuthenticationContext` based on this new setting.

As far as I can tell, AuthServices does not currently process/verify anything in that section today, but by doing it as a compatibility setting we would ease the transition if/when AuthServices does process that section.

AndersAbel:
Novell is wrong. While the section they reference in the spec doesn't mention anything about absolute vs relative URIs there's a general clause in the spec at section 1.3.2:

[...]

PR #1066, opened 2019-02-13
Port IgnoreAuthenticationContextInResponse compatibility option to 2.x

### LINKED ISSUE #1045: The IgnoreAuthenticationContextInResponse compatibility flag not available in version 2.2.0
**Nuget Packages in use:**

- Microsoft.AspNetCore.App - 2.1.1
- Microsoft.NETCore.App - 2.1.0
- Sustainsys.Saml2.AspNetCore2 - 2.2.0

Dotnet Core version: 2.1.302

The IDP incorrectly sets the value of <AuthnContextDeclRef> element to an invalid URI. Due to this following error is generated on SP side while parsing the authentication response:

`error: "IDX13102: Exception thrown while reading 'AuthnContext' for Saml2SecurityToken. Inner exception: 'System.ArgumentException: IDX13300: 'value' must be an absolute Uri, was: 'name/password/uri' at Microsoft.IdentityModel.Tokens.Saml2.Saml2AuthenticationContext.set_DeclarationReference(Uri value) at Microsoft.IdentityModel.Tokens.Saml2.Saml2AuthenticationContext..ctor(Uri classReference, Uri declarationReference) at Microsoft.IdentityModel.Tokens.Saml2.Saml2Serializer.ReadAuthenticationContext(XmlDictionaryReader reader)'."`

The documentation [here](https://media.readthedocs.org/pdf/saml2/latest/saml2.pdf) suggests that Compatibility.IgnoreAuthenticationContextInResponse can be set to true to ignore the `<AuthnContextDeclRef>` element, however, while initiating SPOptions this flag is not available under Compatibility settings.

explunit:
Fixes #1045

See also #780 for the original 1.x feature

### DIFF
--- a/Sustainsys.Saml2/SAML2P/Saml2PSecurityTokenHandler.cs
+++ b/Sustainsys.Saml2/SAML2P/Saml2PSecurityTokenHandler.cs
@@ -18,13 +18,13 @@ namespace Sustainsys.Saml2.Saml2P
 	/// </summary>
 	public class Saml2PSecurityTokenHandler : Saml2SecurityTokenHandler
 	{
+		public Saml2PSecurityTokenHandler(): this(null)
+		{
+			// backward compatibility = null spOptions
+		}
+
 		public Saml2PSecurityTokenHandler(SPOptions spOptions)
 		{
-			if (spOptions == null)
-			{
-				throw new ArgumentNullException(nameof(spOptions));
-			}
-
 			Serializer = new Saml2PSerializer(spOptions);
 		}
 
--- a/Sustainsys.Saml2/SAML2P/Saml2PSecurityTokenHandler.cs
+++ b/Sustainsys.Saml2/SAML2P/Saml2PSecurityTokenHandler.cs
@@ -17,10 +17,10 @@ namespace Sustainsys.Saml2.Saml2P
 	/// could be handled at transport level.
 	/// </summary>
 	public class Saml2PSecurityTokenHandler : Saml2SecurityTokenHandler
-    {
-		public Saml2PSecurityTokenHandler()
+	{
+		public Saml2PSecurityTokenHandler( SPOptions spOptions)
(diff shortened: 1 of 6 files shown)


Sustainsys.Saml2/SAML2P/Saml2PSecurityTokenHandler.cs
```diff
@@ -17,10 +17,15 @@ namespace Sustainsys.Saml2.Saml2P
 	/// could be handled at transport level.
 	/// </summary>
 	public class Saml2PSecurityTokenHandler : Saml2SecurityTokenHandler
-    {
-		public Saml2PSecurityTokenHandler()
+	{
+		public Saml2PSecurityTokenHandler(SPOptions spOptions)
```
AndersAbel: This is a breaking change as the Saml2PSecurityTokenHandler is a public class.

I'd prefer if the old param-less ctor is kept side-by-side and thus the options has to be optional. Which means optional also for the Saml2PSerializer...

What do you think?

Sustainsys.Saml2/SAML2P/Saml2PSecurityTokenHandler.cs
```diff
@@ -17,10 +17,15 @@ namespace Sustainsys.Saml2.Saml2P
...
 		{
-			Serializer = new Saml2PSerializer();
+			if (spOptions == null)
+			{
+				throw new ArgumentNullException(nameof(spOptions));
```
AndersAbel: Missing test coverage on this line - but let's figure out how to deal with the change of signature first.

explunit:
Fixed to allow the default constructor in Saml2PSecurityTokenHandler for backward compatibility. Since Saml2PSerializer is private I did not add a default constructor but I did fix the usage of spOptions to handle null (and added test coverage).

PR #1076, opened 2019-03-26
Mark Saml2 cookie as essential

### LINKED ISSUE #1075: Saml2 cookie is not marked as essential in ASP.NET Core 2.1+
As discussed in https://github.com/Sustainsys/Saml2/issues/666#issuecomment-476652462

### Information needed
1. What nuget packages are you using : Sustainsys.Saml2.AspNetCore2 (2.2.0)
2. What is the expected behaviour : SP-initiated login should work even if the user didn't give their consent for cookies (GDPR consent banner)
3. What happens instead : if cookie consent is enabled and the user didn't give their consent, login doesn't work because the Saml2 cookie is missing (`UnexpectedInResponseToException`)

ASP.NET Core 2.1 introduced support for GDPR cookie consent. This feature is enabled by default in the project template. If it's enabled and the user didn't give their consent, non-essential cookies are not emitted. Since the Saml2 cookie is essential, it should be marked as such.

### Additional info
Please include
* .Net Framework your application is compiled against: `netcoreap2.1`
* .Net Framework installed: .NET Framework 4.7.2, many versions of .NET Core
* Version of Asp.Net MVC / Asp.NET Core used: ASP.NET Core 2.1

thomaslevesque:
Fixes #1075

### DIFF
--- a/Tests/AspNetCore2.Tests/AspNetCore2.Tests.csproj
+++ b/Tests/AspNetCore2.Tests/AspNetCore2.Tests.csproj
@@ -1,13 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.2.0" />
-    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.2" />
-    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.0.1" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.1.2" />
     <PackageReference Include="Microsoft.IdentityModel.Tokens.Saml" Version="5.2.4" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
     <PackageReference Include="MSTest.TestAdapter" Version="1.2.0" />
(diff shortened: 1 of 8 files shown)


Tests/AspNetCore2.Tests/AspNetCore2.Tests.csproj
```diff
@@ -1,13 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
thomaslevesque: This is in order to be able to test the fix. Since the `IsEssential` property doesn't exist in ASP.NET Core 2.0, the `MakeEssential` method would have no effect on that framework.

thomaslevesque:
A test failed during the build, but it's in code not affected by my changes, so I guess it must be a random failure (the test passes on my machine). Amended and force pushed to retry the build.

AndersAbel:
The `Federation_ReloadOfMetadata_KeepsOldDataUntilMetadataBecomesInvalid` is a bad test that fails randomly due to thread pool scheduling timing, just ignore it if it fails. I should really do something about it.

thomaslevesque:
The tests fail because the command to run them specifies "netcoreapp2.0" in the path to the assembly. But since it's not in the source code, I can't change it myself, you'll need to do it in AppVeyor. (If I can make a suggestion, you should consider using an appveyor.yml file instead of specifying everything in the AppVeyor UI. This way build settings are in source control and can be different for each branch if necessary.)

AndersAbel:
Thanks, I'll update the appveyor conf. The reason that the conf is not in appveyor.yml is that I had an integration with Coveralls.io (which is now broken) and that required an API key to be kept secret.

thomaslevesque:
> The reason that the conf is not in appveyor.yml is that I had an integration with Coveralls.io (which is now broken) and that required an API key to be kept secret.

You can keep secrets in appveyor.yml, using secure variables. Or you can set environment variables in the UI so that they're not in the file at all. Anyway, that's a bit off topic ;)

Issue #528, opened 2016-07-19
ID4022: The key needed to decrypt the encrypted security token could not be resolved. Ensure that the SecurityTokenResolver is populated with the required key.


jvdven:
Hi,

We are using Kentor.AuthServices for one of our clients and we are really close to completing our implementation. In one of the last steps an EncryptedTokenDecryptionFailedException is thrown.

When we debug the code we are able to reach line 512 of the CreateClaims method in the Kentor.AuthServices.Saml2P.Saml2Response class (see stacktrace below).

Kentor.AuthServices.Saml2P.Saml2Response.CreateClaims

   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.CreatePlaintextReaderFromEncryptedData(XmlDictionaryReader reader, SecurityTokenResolver serviceTokenResolver, SecurityTokenSerializer keyInfoSerializer, Collection`1 clauses, EncryptingCredentials& encryptingCredentials)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadEncryptedId(XmlReader reader)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadSubjectId(XmlReader reader, String parentElement)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadSubject(XmlReader reader)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadAssertion(XmlReader reader)
   at System.IdentityModel.Tokens.Saml2SecurityTokenHandler.ReadToken(XmlReader reader)
   at Kentor.AuthServices.Saml2P.Saml2Response.<CreateClaims>d__53.MoveNext() in D:\Data\Kentor.AuthServices\Saml2P\Saml2Response.cs:line 512
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Kentor.AuthServices.Saml2P.Saml2Response.GetClaims(IOptions options) in D:\Data\Kentor.AuthServices\Saml2P\Saml2Response.cs:line 476

I'm not sure why there is a key that could not be resolved. Do you know what we should do to resolve this?

Thanks in advance.

m1tchel:
Hi,

[...]

PR #1077, opened 2019-03-27
Add support for EncryptedID

dschilling:
See issue #528.  When an EncryptedID element appears in a SAML response, the deserialization fails with this error:

    NotSupportedException: IDX13140: EncryptedId is not supported. You will need to override ReadEncryptedId and provide support.

... so I have overridden that method, using similar decryption logic to what was already in ReadAssertion.  Should I also bump the version number to 2.3.0?

### DIFF
--- a/Sustainsys.Saml2/SAML2P/Saml2PSerializer.cs
+++ b/Sustainsys.Saml2/SAML2P/Saml2PSerializer.cs
@@ -223,8 +223,18 @@ namespace Sustainsys.Saml2.Saml2P
 					"EncryptedId could not be decrypted using any available decryption certificate");
 			}
 
-			reader = XmlDictionaryReader.CreateDictionaryReader(new XmlNodeReader(decrypted.FirstChild));
+			reader = XmlDictionaryReader.CreateDictionaryReader(new XmlNodeReader(StripWrapper(decrypted)));
 			return ReadNameIdentifier(reader, null);
+		}
+
+		private static XmlNode StripWrapper(XmlElement decrypted)
+		{
+			var inner = decrypted.FirstChild;
+			while (!(inner is XmlElement) && inner != null)
+			{
+				inner = inner.NextSibling;
+			}
+			return inner;
 		}
 
 		public virtual void WriteEncryptedAssertion(XmlWriter writer, Saml2EncryptedAssertion assertion)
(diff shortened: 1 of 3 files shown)


AndersAbel:
Thanks! I'll have a look and merge it. Please don't touch the version number - I'll bump that when I release.

RaptorCZ:
I need this in release version. Any info about merge to master?

AndersAbel:
> I need this in release version. Any info about merge to master?
> 
> And just a question - how to "activate" it in SPOptions? Tried to build custom version with pull request changes but getting error
> 
> ```
> IDX13117: A <saml:EncryptedAttribute> was encountered while processing the attribute statement.To handle encrypted attributes, extend the Saml2SecurityTokenHandler and override ReadAttributeStatement.
> ```
> 
> What to do?

EncryptedAttribute and EncryptedId are two different things. Are you sure that this is the PR you need?

A better option would probably be to help out on the v3 version where we can be more flexible and use functionality from Microsoft.IdentityModel that looks to be due for .NET Core 5. See AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet#734

RaptorCZ:
> EncryptedAttribute and EncryptedId are two different things. Are you sure that this is the PR you need?

With version 2.2 I've got error 
```
IDX13140: EncryptedId is not supported. You will need to override ReadEncryptedId and provide support.
```
So I got this PR to get support for `EncryptedId`. But now I have another error listed in previous post.
```
IDX13117: A <saml:EncryptedAttribute> was encountered while processing the attribute statement.To handle encrypted attributes, extend the Saml2SecurityTokenHandler and override ReadAttributeStatement.
```
Error is thrown in file `Saml2PSerializer` in code
```c#
if (envelopeReader.IsStartElement(Saml2Constants.Elements.AttributeStatement, Saml2Constants.Namespace))
						statement = ReadAttributeStatement(envelopeReader);
```

I just need to be able process customer's SAML data and authenticate user, in current full .net framework application.

AndersAbel:
@RaptorCZ Looks like this PR is not the only thing you need then. There are several different options for encryption in a Saml message, so I think this would need some rework.

And that kind of rework is nothing I am ready invest time in on the v2 branch. Actually, making the design more flexible and adding a decryption step to the parsing is one of the reasons I want to design a v3.

Is this something you would want to invest time in coding on? Or is there an organisation that needs this that would be interested in sponsoring the development work needed to get full EncryptedId/Assertion/Attribute support?

PR #1090, opened 2019-05-08
Update .gitignore

marlinspike:


AndersAbel:
Why include bin in source control?

PR #1096, opened 2019-05-30
CSP hash support for HttpPost binding

### LINKED ISSUE #1093: Content Security Policy (CSP) support
When HTTP posting to the SAML IDP, the HTML is generated by a constant in the code and returned to the browser. The javascript in this HTML is inline and requires the CSP to have 'unsafe-inline' enabled. This is a security concern. It would be nice to at least specify a nonce for the script or to allow the script to be in a separate javascript file.

explunit:
Fixes #1093 by moving the auto-posting script into its own script tag. Browsers do not support hash for inline events.

Also updated SampleMvc project to use NWebSec to emit the CSP header so this could be tested during development. During this process I removed Modernizr since it didn't seem to be used and was causing difficulty with getting clean CSP results.

The example is slightly wrong in that it allows any page (not just the /SignIn route) to host a script that does `document.forms[0].submit();`. But I think the architecture for CSP is outside the scope of what we normally try to provide in sample applications.


fuzzzerd:
General CSPs allow loading from same origin, so modifying the bundles is not necessary. I believe the issue is with the inline script.

I think simply changing this: 
```
<body onload=""document.forms[0].submit()"">
```
to this:
```
<meta http-equiv="Content-Security-Policy"
        content="script-src 'sha256-8lDeP0UDwCO6/RhblgeH/ctdBzjVpJxrXizsnIk3cEQ='">
<script type=""text/javascript"">
document.forms[0].submit();
</script>
```
I think I've seen somewhere that the hash can be on the script tag, but I can't find an example of that to reference at the moment.

AndersAbel:
I think that allowing any page to contain `documents.forms[0].Submit();` is too wide. I can see how that could be used to circumvent some cross site request forgery protections. For example to call the log out button from cross site scripting as part of a session pinning attack. I think that the approach by @fuzzzerd looks better as it only affects our page.

explunit:
@fuzzzerd @AndersAbel I've changed to also emit the CSP hash in a meta tag. The problem, of course, is that CSP in the headers overrides CSP in the meta tags. Most people who are implementing a [strict CSP](https://csp.withgoogle.com/docs/strict-csp.html) will have a global handler (nwebsec or otherwise) to emit headers site-wide. So they will need to do something similar to what I did in the (somewhat bogus) SampleMvc and customize the output for this particular route to include the hash for this script.

Having the hash in the meta tag is unlikely to have much real world utility for the developer, except that:
1 - it provides an easy place to find the hash so they can include it in their headers config for this route
2 - for someone without a global CSP, it provides some incremental protection against a future XSS problem discovered in this SignIn HttpPost binding html.

fuzzzerd:
That is a good point. The meta tag can only make the CSP more strict, not less. For cases where no other CSP header is already being sent, this is a step up. For cases when there is a CSP header, this gives consumers of the package an easy place to grab the hash they need to include in their header.

AndersAbel:
If the hash will be transmitted globally, I think the script should be adjusted. Using `documents.forms[0].Submit();` might be a CSRF attack vector. It would be better to give the form a proper id and reference it by id instead. That would reduce the attack to forms having the same id.

explunit:
@AndersAbel fixed in 1f9e359. The hash wasn't transmitted globally in our sample, but you're right that using a named form makes it easier for developers to avoid that mistake in their own applications.

PR #1105, opened 2019-07-19
Fixed configoption name outboundSigningAlgorithm

devince001:
outboundSignAlgorithm doesn't match the name in de code


AndersAbel:
Thanks for noticing and taking the time to fix!

PR #1122, opened 2019-10-29
Do not modify shared options in Notifications test

explunit:
These tests were modifying a shared options object and causing indeterminate results based on concurrent running of other tests. I'm not sure what changed (.Net 4.8? Visual Studio 2019?) to make them fail now but not earlier.


explunit:
Looks like AppVeyor is also happy with these changes, so I'll plan to merge in next day or two unless there are any objections.

AndersAbel:
Good catch. When development started I used shared options, but then moved away from it. Obviously some old bad habits still are around in the code.

PR #1131, opened 2019-12-07
AspNetCore2: CommandFactory as a service, Command methods to virtual, some extension classes to public

jnivala:
These non-breaking changes allow for convenient overriding of default command implementations.

With these changes, you can use add your own ICommandFactory as a service to instantiate e.g. your own Acs command. Command methods are changed to virtual. That allows you to inherit default command implementations and selectively override their methods.

Some extension classes are changed to public to allow you to reuse them.

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/CommandResultExtensions.cs
+++ b/Sustainsys.Saml2.AspNetCore2/CommandResultExtensions.cs
@@ -8,8 +8,20 @@ using System.Threading.Tasks;
 
 namespace Sustainsys.Saml2.AspNetCore2
 {
-    static class CommandResultExtensions
+    /// <summary>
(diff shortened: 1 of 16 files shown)


jnivala:
I just discovered Options.Notifications. But it seems they do not provide enough control over behavior: E.g. how do you override default Acs command behavior to allow graceful handling of IDP returning a login error? Allowing user to have control over command instantiation would provide that.

Overall, it would be nice to see more DI used. I'm thinking about Identityserver4: it is very configurable as it allows a lot of default implementations overridden via DI.

AndersAbel:
I won't merge this as is. There are ideas about a version 3.X that would use DI extensively and probably only target Asp.Net Core. Getting rid of the HttpModule and Owin middleware would make things MUCH easier.

PR #1133, opened 2019-12-11
Support for IsPassive added to SAML request

GastonCZ:
I have added support for the IsPassive flag. It is read from the API call and added to the SAML request. The parsing of the SAML "nopassive" flag that can be a response from the IDP on the passive logon request was already implemented.


GastonCZ:
This pull request has to wait. I implemented IsPassive the same way as ForceAuthn was already implemented. During testing I found out that this is not enough. The IsPassive and ForceAuthn flags are not parsed from the input so they are never set in the SAML output. I am working to fix this and will update the pull request.

GastonCZ:
I have added support for IsPassive and ForceAuthn to the parsing of the request. Also fixed the unit tests that failed due to my changes.

PR #1135, opened 2019-12-16
Set the cookie "Secure" flag iff the ACS post back URL is HTTPS

### LINKED ISSUE #1091: Please support SameSite=None and Secure when setting cookies
Google announced that they will soon be requiring a new explicit SameSite=None property in order to send cookies on cross site requests. 

https://blog.chromium.org/2019/05/improving-privacy-and-security-on-web.html
https://web.dev/samesite-cookies-explained/

With Chrome 76, as described in the second link, with the `same-site-by-default-cookies` and  `cookies-without-same-site-must-be-secure` flags set, cookies will no longer be sent when the IdP POSTs to the ACS unless `SameSite=None; Secure` was set on the cookie, which will break the current implementation.

Please support setting this new header so that sites using Sustainsys do not break when this change arrives.

Related, I've noticed that in ASP.NET Core that setting SameSiteAttribute.None does not explicitly send a `SameSite: None` cookie attribute as that was assumed by the implementation and previous specification to be the default. I worked around this by manually setting the cookie with 

```
response.Headers.Append("set-cookie", $"{key}={value}; path=/; SameSite=None; Secure");
```

Similar workarounds will probably be necessary for other framework implementations.

### Related issues
This is the main issue for tracking the SameSite updates. There are also separate issues for each API package on the 1.X and 2.X versions. Please see list of related issues below.

ulrichb:
Partially fixes #1091 (for the ASP.NET _Core_ version).

This allows to comply with the new (Chrome >= 80) "Reject insecure SameSite=None cookies" rule (which otherwise would drop the `SignInCommand` correlation cookie).

Manually tested: 
 * ASP.NET Core: Chrome 80 (Dev Channel) with enabled "SameSite by default cookies" and "Cookies without SameSite must be secure" flags in chrome://flags/

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/CommandResultExtensions.cs
+++ b/Sustainsys.Saml2.AspNetCore2/CommandResultExtensions.cs
@@ -35,6 +35,7 @@ namespace Sustainsys.Saml2.AspNetCore2
                     new CookieOptions()
                     {
                         HttpOnly = true,
+                        Secure = commandResult.SetCookieSecureFlag,
                         // We are expecting a different site to POST back to us,
                         // so the ASP.Net Core default of Lax is not appropriate in this case
                         SameSite = SameSiteMode.None,
--- a/Sustainsys.Saml2.HttpModule/CommandResultHttpExtensionsShared.cs
+++ b/Sustainsys.Saml2.HttpModule/CommandResultHttpExtensionsShared.cs
@@ -42,7 +42,8 @@ namespace Sustainsys.Saml2.HttpModule
                     commandResult.SetCookieName,
                     protectedData)
                 {
-                    HttpOnly = true
+                    HttpOnly = true,
+                    Secure = commandResult.SetCookieSecureFlag,
                 });
             }
 
--- a/Sustainsys.Saml2.Owin/CommandResultExtensions.cs
+++ b/Sustainsys.Saml2.Owin/CommandResultExtensions.cs
@@ -71,6 +71,7 @@ namespace Sustainsys.Saml2.Owin
                     new CookieOptions()
                     {
                         HttpOnly = true,
+                        Secure = commandResult.SetCookieSecureFlag,
                     });
             }
 
--- a/Sustainsys.Saml2/WebSSO/CommandResult.cs
+++ b/Sustainsys.Saml2/WebSSO/CommandResult.cs
@@ -68,6 +68,11 @@ namespace Sustainsys.Saml2.WebSso
         public string SetCookieName { get; set; }
 
         /// <summary>
+        /// Value of the "Secure" flag for the cookie (relevant if <see cref="SetCookieName"/> != null).
+        /// </summary>
+        public bool SetCookieSecureFlag { get; set; }
+
+        /// <summary>
         /// SAML RelayState value
         /// </summary>
         public string RelayState { get; set; }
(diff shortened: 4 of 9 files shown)


ulrichb:
@AndersAbel This is a proposal for https://github.com/Sustainsys/Saml2/issues/1091#issuecomment-565257634.

I use the `AssertionConsumerServiceUrl.Scheme` because it's exactly the post back URL and also includes the `PublicOrigin` (if set).

What do you think?

ulrichb:
@AndersAbel Could you take a look at this? Chrome 80 will be released in February ....

AndersAbel:
@ulrichb I had a quick look and I think it looks good, but I need to look a bit more in detail on it before merging.

Is this the only change needed for Chrome 80 to work? I assumed that a SameSite value must also be set (except for that old version of Safari that requires agent sniffing to detect)?

ulrichb:
> Is this the only change needed for Chrome 80 to work?

I successfully tested the _login_ procedure (against an ADFS Server) with Chrome 80 Dev Channel (+ enabled _"SameSite by default cookies"_ and _"Cookies without SameSite must be secure"_ in chrome://flags/).
(Did not test the logout federation.)

---

>  I assumed that a SameSite value must also be set

For ASP.NET Core SameSite=None is already set. (See `CommandResultExtensions`).

For `Sustainsys.Saml2.Mvc`, a target framework change to 4.7.2 would be necessary to set SameSite = None. See https://docs.microsoft.com/en-us/dotnet/api/system.web.httpcookie.samesite.

For Owin, a library update to 4.1.0 seems to be necessary. See https://github.com/aspnet/AspNetKatana/issues/201.

(I only use the ASP.NET Core version.)

ulrichb:
(I updated the description above to "_Partially_ fixes #1091 (for the ASP.NET _Core_ version).")

mklinke:
@ulrichb What version of ASP.NET Core have you tested against? Also, won't this break local debugging without HTTPS? At least that's possible with the current implementation. Not sure if the change in Chrome 80 alone will already break that...

ulrichb:
ASP.NET Core 2.2. Oh, I forgot to mention that the explicit `SameSite=None` emitting only happens when you have ASP.NET Core >= 2.1.14 or ASP.NET Core >= 2.2.8, before these patch versions the "None" value was simply ignored (see https://github.com/aspnet/AspNetCore/pull/13746.)

---

> Also, won't this break local debugging without HTTPS?

That's the reason why I used `urls.AssertionConsumerServiceUrl.Scheme` ... to avoid the secure flag on HTTP hosts (or theoretically PublicOrigin's). But yes, HTTP localhost testing doesn't work in Chrome >= 80 (unless you explicitly disable the flags, which you can do for debugging purposes); but this is a general problem for all non-HTTPS (localhost debugging) cross origin cookies ...

ulrichb:
@mklinke Do you need any further info? (February is coming ....)

mklinke:
@ulrichb Not sure, what you're asking. I'm yet another user of the library as of now. My project is currently based on .NET Core 3.0 and I found the issue as a blocker when I tried upgrading to .NET Core 3.1 due to the SameSite breaking change there.

ulrichb:
Sorry, I meant @AndersAbel :)

AndersAbel:
@ulrichb @mklinke I'm working on the releases for this right now, please see #1091 for status as that's the main issue.

ulrichb:
Thanks @AndersAbel !

AndersAbel:
Note: Need to apply the Secure flag for the redirect to Discovery Service too, I'll fix that when merging.

ulrichb:
Cool. Thx!

AndersAbel:
I ended using cherry-pick to integrate this - first to the v1 branch and then to master. So I'm marking this PR as "closed" but it means "merged"

ulrichb:
@AndersAbel Many thanks!

PR #1158, opened 2020-02-13
Create xml modifications

therrealdude:
I figured it out. Let me know if this needs any more work. This is time sensitive for the client I am working with, but I'm not really sure what it takes to push a new version of the nuget package so do what you can.

### DIFF
--- a/Sustainsys.Saml2/SAML2P/Saml2RequestBase.cs
+++ b/Sustainsys.Saml2/SAML2P/Saml2RequestBase.cs
@@ -177,6 +177,11 @@ namespace Sustainsys.Saml2.Saml2P
         public abstract string ToXml();
 
         /// <summary>
+        /// Event that fires just before xml is returned.
+        /// </summary>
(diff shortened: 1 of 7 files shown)


therrealdude:
@AndersAbel , I created a pull request for my issue here. Let me know if there is anything else that would be needed for this.

Sustainsys.Saml2/SAML2P/Saml2RequestBase.cs
```diff
@@ -176,6 +176,11 @@ private void ValidateCorrectDocument(XmlElement xml)
         /// <returns>string containing the Xml data.</returns>
         public abstract string ToXml();
 
+        /// <summary>
+        /// Event that fires just before xml is returned.
+        /// </summary>
+        public Action<XElement> XmlCreated = delegate { };
```
AndersAbel: The callback should be set in the Saml2Notifications class.

Issue #1150, opened 2020-01-17
Use CookieManager in Asp.Net Core


AndersAbel:
Asp.Net Core uses a Cookie Manager that can intercept and alter cookies that are about to be set. The Sustainsys.Saml2 way to handle that is a notification that decided when to emit a SameSite=None header or not based on the user agent string. It would be good to offer the standard cookie manager for use with the Asp.Net Core module too.

Intention was to get this into the 2.4.0 version with the SameSite fix #1091. Unfortunately it didn't make it due to being pressed of time.

If this is a breaking issue for anyone, please comment - it might be possible to make another small release with this implemented.

explunit:
`Microsoft.Owin.Infrastructure.ICookieManager` would be very useful for the OWIN mode too, for things like #594 which continues to cause problems for us.

PR #1170, opened 2020-03-18
Support ICookieManager in ASP.NET Core

explunit:
The ASP.NET Core version of #1150

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/Saml2Options.cs
+++ b/Sustainsys.Saml2.AspNetCore2/Saml2Options.cs
@@ -21,8 +21,6 @@ namespace Sustainsys.Saml2.AspNetCore2
             {
                 ModulePath = "/Saml2"
             };
-            // uses the headers directly when no chunk size is set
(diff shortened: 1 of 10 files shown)


Sustainsys.Saml2.AspNetCore2/Saml2Options.cs
```diff
@@ -20,6 +21,8 @@ public Saml2Options()
             {
                 ModulePath = "/Saml2"
             };
+            // uses the headers directly when no chunk size is set
+            CookieManager = new ChunkingCookieManager();
```
AndersAbel: The default value can be set in PostConfigureOptions instead, to avoid creating an object that is then over-written by custom config:
https://github.com/Sustainsys/Saml2/blob/0859e67392ea613fcb82a8d7e0e9a6c1c4de6ab6/Sustainsys.Saml2.AspNetCore2/PostConfigureSaml2Options.cs#L31

AndersAbel:
Marking as WIP because reading should also be done through the Cookie Manager before merge.

PR #1178, opened 2020-03-25
feat: Add optional compatibility option to support non-complaint IDPs that do not sign their LogOut responses.

henrikottesorensen:
We got hit by issue https://github.com/Sustainsys/Saml2/issues/446, after our IDP began to returning unsigned LogOut responses. I have added a compatibility switch to support this specification violating behaviour, while we work with the IDP provider to get them in compliance with the spec.

### DIFF
--- a/Sustainsys.Saml2/Configuration/Compatibility.cs
+++ b/Sustainsys.Saml2/Configuration/Compatibility.cs
@@ -80,5 +80,13 @@ namespace Sustainsys.Saml2.Configuration
         /// </summary>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1726:UsePreferredTerms", MessageId = "Logout")]
         public bool EnableLogoutOverPost { get; set; }
+
+        /// <summary>
+        /// SAML2 Specs says in section 4.4.4.2:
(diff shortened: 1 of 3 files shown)


AndersAbel:
I'm a bit confused about what the scope is of this PR. The title mentions "Logout Responses". The code excludes both logout responses and incoming logout requests from the signature check. The test case is for a logout request.

Do you want this to apply only to logout responses? Or only to requests? Or both?

From a security perspective I can't think of why a missing signature on a logout response is a problem. A missing signature/sender validation for an incoming logout request however might be part of an attack vector. If we compare to OpenID Connect they authenticate incoming logout requests (by including the id_token) but not the redirect back.

henrikottesorensen:
My apologies for the late reply, been putting out various brush fires and then easter holiday hit us.

Being unfamilar with the Saml2 protocol and the codebase, so the overly aggressive patch is my fault.
 
For our use case, until the IDP becomes spec complient, we only need to support unsigned logout **RESPONSES**. I can supply you with an example HAR archive of the traffic, if needed.

AndersAbel:
@henrikottesorensen Thank you for your response. I will go ahead and merge this with a modification to only allow unsigned logout responses.

PR #1182, opened 2020-04-10
Break out Metadata to separate library

rpmansion:
Move metadata codes from Sustainsys.Saml2 to Sustainsys.Saml2.Metadata project except for codes related to loader and scheduler.


rpmansion:
@AndersAbel should we use the `EntityId` from `IdentityModel`? If yes, will remove the one in the code base.

AndersAbel:
> @AndersAbel should we use the `EntityId` from `IdentityModel`? If yes, will remove the one in the code base.

What assembly is it in?

rpmansion:
> What assembly is it in?

System.IdentityModel

AndersAbel:
> > What assembly is it in?
> 
> There is one in `System.IdentityModel` and one in `Sustainsys.Saml2.Metadata`

As far as I know the one in `System.IdentityModel` is not available in .NET Core. That's the reason we created an own one.

rpmansion:
> As far as I know the one in System.IdentityModel is not available in .NET Core. That's the reason we created an own one.

I see. Upon checking again, still not available in .NET Core

rpmansion:
@AndersAbel need to rerun the action, I'm getting all of the test as passed on my local

AndersAbel:
Ignore it, the report says "internal error". I hope the next commit will work.

rpmansion:
@AndersAbel I've moved codes of the metadata except the loader and scheduler

AndersAbel:
@rpmansion Sounds great. I'll have a look and get it merged.

rpmansion:
@AndersAbel reviewing the code base, what is the reason why we do not maximize the use of `System.Security.Cryptography.Xml` instead of creating our own classes that are already exist in such library?

AndersAbel:
> @AndersAbel reviewing the code base, what is the reason why we do not maximize the use of `System.Security.Cryptography.Xml` instead of creating our own classes that are already exist in such library?

Do you have an example? I can think of two reasons: 1. It was not available on .NET Core when the port was done. 2. It doesn't contain all the properties to properly represent the metadata.

AndersAbel:
Good first step to breaking metadata out to a separate library. Thanks!

rpmansion:
@AndersAbel can we link this to the issue? Thank you for merging, will further add new changes. :)

PR #1219, opened 2020-06-25
Bug/1062 ignore authentication context in response not available in web.config

### LINKED ISSUE #1062: IgnoreAuthenticationContextInResponse not available in web.config (1.x)
The IgnoreAuthenticationContextInResponse compatibility flag cannot be set through web.config, only in code. Should be added to web.config.

madjufra:
Fixes #1062


AndersAbel:
web.config is going out of support, there will be no more features added to that.

PR #1236, opened 2020-08-28
Adding fips compliance changes to Sustainsys library

aberro3s:
-Changed CryptographicExtensions class to use AesCryptoServiceProvider instance for FIPS valid algorithm.
-Changed and override decrypt methods from .Net on RSAEncryptedXML class for creating fips algorithm valid instances, avoid using CreateFromName method.
-Added Test methods for RSAEncryptedXML class and CryptographicExtensions as well as a helper class for test initialization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,11 @@ There are three active branches in the repo
 
 ## Documentation
 Complete documentation is available at [our documentation site](https://saml2.sustainsys.com).
+
+##Fips Compliance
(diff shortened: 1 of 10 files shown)


AndersAbel:
Please have a look at the new signature validation code that's part of the Metadata library in `develop`. I think that when it uses the certificates non-touched, it should be possible to run on a FIPS-compliant platform.

PR #1241, opened 2020-09-15
Add test cases for documented XML Signature Wrapping exploits

cmunky:
This PR adds test cases for documented XML Signature Wrapping exploits

During the course of our SAML security audit, we came across a [blog post identifying several specific XSW exploits](https://blog.netspi.com/attacking-sso-common-saml-vulnerabilities-ways-find/).

We followed up on this and attempted each exploit ourselves [using available tools](https://portswigger.net/bappstore/c61cfa893bb14db4b01775554f7b802e).

As result of this work, we wrote test cases to ensure these exploits continue to return the expected exceptions

We based these test cases on the [libraries provided by the SAML Raider extension](https://github.com/PortSwigger/saml-raider/blob/master/src/main/java/helpers/XSWHelpers.java)

We felt that these test cases may also be beneficial to other users of the Sustainsys SAML libraries


mapitman:
Any chance of getting this merged?

AndersAbel:
This is the kind of PR that gets an OSS maintainer very very happy.

Issue #1238, opened 2020-09-04
AES-GCM support


jnivala:
We are facing a requirement to support receiving SAML2 responses with https://www.w3.org/2008/xmlsec/namespaces.html#aes128-gcm encrypted assertions on .NET 4.7.2 and .NET Core 3.1.

NET 4.7.2 does not contain the algorithm which leaves us with BouncyCastle implementation.

.NET Core 3.1 has it, but EncryptedXml, that handles decrypting, seems to be tightly coupled with SymmetricAlgorithm class that the AES-GCM implementation does not inherit from. So it seems you can not just plug in AES-GCM support even if framework contained it.

Any thoughts about this?

Regards.

jnivala:
Ok, for 4.7.2 where we construct Saml2Response manually it seems we would need:
- Handle decrypting XML element inside Saml2Response rather than in extension class.
- Add a method with "public virtual EncryptedXml GetEncryptedXml(AsymmetricAlgorithm key, XmlDocument xmlDoc)" signature to Saml2Response to allow inheriting class to decide the EncryptedXml implementation.
- Open RSAEncryptedXml as public.

Then we can extend RSAEncryptedXml with our own decryption code and tell our Saml2Response implementation to use it.

For .NET Core where we use the ASP NET Core pipeline, it would be the previous plus:
- To AcsCommand add method "public virtual Saml2Response GetSaml2Response(HttpRequestData request, IOptions options, UnbindResult unbindResult)" to allow decide Saml2Response implementation.
- Change AcsCommand.ProcessResponse from private static to public virtual.
- Open CommandResultExtensions class as public to allow Apply mehtod from used from outside.
- Make Saml2Handler public methods virtual.

=> It seems we neet to fork.

JukkaM:
We are facing the same requirement as well. It would be great if we could have proper converstation here to decide how to proceed. @jnivala pointed out some refactorings which would allow us to extend the library so that we could build AES-GCM support on our own but is there plans to add AES-GCM support inside the library? Which is the best practice you suggest?

Naturally extensibility is great thing to support but adding more features to the library itself is good thing as well.

AndersAbel:
The idea is to make things more extensible i v3 - yes.

And if the tight coupling with SymmetricAlgorithm can be removed it might be an idea. It is also worth checking out what features are available in Microsoft.IdentityModel. They have their own XML parser that might support these scenarios.

jaleikas:
[...]

PR #1251, opened 2020-11-05
Enable adding support for AES-GCM 128

laurihelkkula:
For #1238 , enable adding support for AES-GCM (http://www.w3.org/2009/xmlenc11#aes128-gcm) in incoming SAML messages. Add sample project.

### DIFF
--- a/Samples/SampleAesGcmStartupNetCore31/AesGcmAlgorithm.cs
+++ b/Samples/SampleAesGcmStartupNetCore31/AesGcmAlgorithm.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Security.Cryptography;
+
+namespace SampleAesGcmNetCore31
+{
+  /// <summary>
+  /// SymmetricAlgorithm decrypting implementation for http://www.w3.org/2009/xmlenc11#aes128-gcm.
+  /// This is class is not a general implementation and can only do decryption.
+  /// </summary>
+  public class AesGcmAlgorithm : SymmetricAlgorithm
+  {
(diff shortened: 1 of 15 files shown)


mronnlun:
@AndersAbel @laurihelkkula Are there any more examples on how to use this code?

I'm trying to use EncryptedXml.DecryptData but I'm struggling to supply an instance of SymmetricAlgorithm. I have a pfx file and it's password.

Could you be so kind as to elaborate on the example, e.g. by completing a method like the one here:

```
public string Validate(string SAMLResponse)
{
  var xml = Encoding.UTF8.GetString(Convert.FromBase64String(SAMLResponse));

  var xmlDoc = new XmlDocument();
  xmlDoc.PreserveWhitespace = true;
  xmlDoc.LoadXml(token);

  var encryptedElement = xmlDoc.GetElementsByTagName("xenc:EncryptedData")[0] as XmlElement;
  var encryptedData = new EncryptedData();
  encryptedData.LoadXml(encryptedElement);

  var encryptedDoc = new EncryptedXml(rawResponse);
  encryptedDoc.DecryptData(encryptedData,  xxx);
}
```

laurihelkkula:
I'm not sure how your question relates to this PR, or am I missing something? If you want to use the EncryptedXml class directly, the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.xml.encryptedxml?view=net-5.0) have some examples of how to use the EncryptedXml class.

AndersAbel:
@mronnlun 1. That code doesn't look at all that something that is related to this library. 2. A merged pull request is not the right place to add new questions. If this truly is about using the `Sustainsys.Saml2` library with AES-CGM, then try opening a new issue.

PR #1254, opened 2020-11-13
Fixes #1252

creste:
Added support for detecting .NET 5 at runtime.

PR #1255, opened 2020-11-14
Refactor EnvironmentHelpers.IsNetCore checks

### LINKED ISSUE #1252: .NET 5: EnvironmentHelpers.IsNetCore bad implementation
Hi,
implementation of flag `EnvironmentHelpers.IsNetCore` in `Saml2.Internal` and `Saml2.Metadata` isn't good.
If main application is in .net 5, detection fails and library crashes. 

After forking project and fixing bug, library starts to working properly.

dlemstra:
This PR removes the `EnvironmentHelpers.IsNetCore` checks:
- The cast to `RSACryptoServiceProvider` is only required when the library is targeting the .NET framework. That has been changed into a compile time check.
- The check for the available signing and digest algorithms is no longer required after upgrading `System.Security.Cryptography.Xml` to `5.0.0`.

To test if the library works I localy added a Tests.NET5 project. To get all tests passing I had to make a small change to one of the unit tests because the exception message has changed in .NET 5.0. That project should probably be added in a separate PR because that will probably require some changes in the build. I can open up a new PR for that after this has been merged.

This PR also contains some additional changes:
- Removed hard coded xml document paths.
- Marked the `Sustainsys.Saml2.Metadata` assembly as `CLSCompliant` to silence the compiler warnings.

Fixes #1252


dlemstra:
Also just pushed a change to re-enable the GitHub workflow.

AndersAbel:
There are a lot of changes in here that are not only related to .NET 5. I intend to pick what is critically important into the v2 branch to get a release out. Are the small changes in #1254 enough? /cc  @creste

dlemstra:
All these changes are related to the .NET 5 compatibility. It removes all the .NET core checks. The other PR is a simple quick hack to make it work now.

AndersAbel:
@dlemstra Thanks. So would it make sense to get the other PR into v2 for a quick release now and get these changes into develop?

dlemstra:
Why would you not want to merge this one instead of the hack? I can also make a PR for the V2 branch if that would help you?

AndersAbel:
This one requires more time to review. The hack is easier to understand. And for v2 I just want to get something out quickly.

For v2 it is also important that the code still runs on .NET Core 2, .NET 3 and the full NET Framework. Does your PR keep that? Or does it make the library .NET 5 only?

dlemstra:
The library currently targets the folowing frameworks: `netstandard2.0;net47;net461`. There are two unit tests projects one for .NET Core that uses the code in the netstandard2.0 library and the NETFramework one uses the net47 version of the library. Because this removes all the run time checks for the platform these tests should guarantee that the library works for  .NET Core 2, .NET 3 and the full NET Framework.

It might be an option to also include a test project for `netcoreapp3.1` and `net50` to also run these tests there. I have already set this up localy for `net50` to double check that my changes made the library work with `net50`.

AndersAbel:
I've looked through this a bit more now and am starting to understand how it works. It looks like a good refactoring base don the structure in the `develop` branch. Although I am not happy with having changes to the `dotnetcore.yml` or removal of XML documentation files in the same PR.

But for the v2 that I just want to get shipped with minimum rework it is probably easier to just go with the hack in #1254.

The thing is that the v2 design/architecture is past what is productive to work with. But there can be no breaking changes in there. So I think rethinking the design in `develop` for v3 is reasonable, but sticking with smallest possible fix in v2.

dlemstra:
Just dit a quick test to see if my `net50` project would pass all tests in the v2 branch and they all pass expect the argument null check test that was changed in this PR. I can split this PR up into smaller pieces if that would be more convenient.

dlemstra:
Maybe  #1307 will finally get these changes merged. Going to close this now.

Issue #1270, opened 2021-01-28
SPSSODescriptors aren't ignored on metadata parsing


BenediktHeinrichs:
In the issue https://github.com/Sustainsys/Saml2/issues/408 the functionality of creating SPSSODescriptors was excluded since metadata from federations sometimes include SPs which have duplicate index numbers on their AssertionConsumerServices.

This however was "removed" in https://github.com/Sustainsys/Saml2/commit/535c5703bc8385a1dabdb7742411493bd2be352c#diff-73db87bfa67a498d11677ba605dea2195bc0d42eed151ae28de5498255e69d69R59 and since then this does not work anymore this way.

Since we stumbled upon this issue while including the DFN-AAI federation, this issue proposes to reenable the previous functionality and to adapt a previous test which should deal with checking this.

BenediktHeinrichs:
Added two merge request, so that this can be merged into [develop (v3)](https://github.com/Sustainsys/Saml2/pull/1271) for future usage and [v2](https://github.com/Sustainsys/Saml2/pull/1272) for direct adoption.

AndersAbel:
The new metadata reader that's now in `develop` contains functionality to ignore/supress errors, so this won't be needed.

PR #1272, opened 2021-01-28
Re-Add the functionality to exclude SpSso creation

BenediktHeinrichs:
This pull request deals with #1270 in v2.

The functionality for excluding the creation of SpSsoDescriptors is reenabled and migrated according to the new naming structure.
A test is altered to the state it was initially in on creation, for dealing with the linked issue.


AndersAbel:
Won't be fixed in `v2`, `v3/develop` contains a new, more resilient metadata parser.

Issue #570, opened 2016-09-29
[Question] Case-sensitive comparison of the Request.Path and the acsPath strings - why?


dilyandimitrov:
Hi guys, 

Have a look here - 
https://github.com/KentorIT/authservices/blob/aspnetcore/Kentor.AuthServices.AspNetCore/KentorAuthServicesHandler.cs#L22

Is this made like that by purpose and why? What is the reason this comparison not to be case-insensitive ?

Dilyan

AndersAbel:
@mrfatgoat The reason that this is case sensitive is that i just happend to, there's no thought behind it. I think that we should follow whatever case settings that Microsoft's handlers for Owin and ASP.NET Core use. Anyone knows (or can check?)

mrfatgoat:
Ok, I chased down some answers to this.  I thought Microsoft's OpenId connect authentication handler would be a pretty good analogy for the AuthServices authentication handler.

Starting here in [OpenIdConnectExtensions](https://github.com/aspnet/Security/blob/rel/2.0.0/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectExtensions.cs#L26), we see the handler class [OpenIdConnectHandler](https://github.com/aspnet/Security/blob/rel/2.0.0/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L30) is used, which derives from [RemoteAuthenticationHandler<TOptions>](https://github.com/aspnet/Security/blob/rel/2.0.0/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs).  `OpenIdConnectHandler` overrides `public Task<bool> HandleRequestAsync()` [here](https://github.com/aspnet/Security/blob/rel/2.0.0/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L76) and performs comparisons between `PathString` objects using the `==` operator.  The [PathString == operator](https://github.com/aspnet/HttpAbstractions/blob/rel/2.0.0/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs#L379) calls `Equals()` which calls `Equals(other, StringComparison.OrdinalIgnoreCase)`.

[...]

PR #1280, opened 2021-03-20
Case insensitivity on ModulePath

ghost1face:
This is a pull request targeting branch `v2` for Framework & Core2 implementations.  

This resolves the issue of the Saml2 handler only listening for requests that start with the **case sensitive** module path.  This allows for case insensitivity resolving #570 

This COULD potentially be breaking if current consumers rely on the path to be exact case.  If this is a large concern, I'm open to providing a parameter on `SPOptions` which defines if comparisons should be sensitive or not with a default of case-sensitive comparisons


AndersAbel:
Closing PRs from before the start-over in `develop` as they do not make sense when the entire library is rewritten.

PR #1299, opened 2021-06-30
Allow external usage of SPOptionsExtensions

Tecknoth:
This would allow metadatas to be generated outside of the package scope.


AndersAbel:
Improved metadata generation is part of what's now in `develop` for `v3`.

Issue #896, opened 2018-01-24
Change HttpRequestData into HttpRequestData<T>


AndersAbel:
Add a templated type member to HttpRequestData that contains the HttpContext, owin context or asp.net core HttpContext that it was created from. Reason is to be able to extract context inside the notifications.

AndersAbel:
As discussed in #895

PR #1310, opened 2021-09-22
Add HttpRequestData<T> for notification context

toddlucas:
This addresses Sustainsys/Saml2#896.

Given that the core library is still targeting .NET Standard, we can't introduce .NET Core's `HttpContext` yet. As a bridge to that point, this solution derives a class with a generic parameter from the current `HttpRequestData`, which has a new object property to hold the context. This is a similar approach to `IEnumerator<T>` vs. the object-based `IEnumerator`. Although in that case, the `Current` property is suppressed with the `new` modifier, whereas this solution uses a virtual property with a second typed property in the derived class, so that the instance data can be shared instead of being duplicated.

The new filename uses a name based on the .NET team's naming conventions. Happy to change it.

The notification handler signatures have been modified to take the new context object. Implementors will need to cast to their target-specific type (i.e., `HttpContext` on Core or `HttpRequestBase` on Framework). This can be replaced with a typed parameter if/when this library becomes Core only.

Only two notifications were updated--those required to support loading `IdentityProvider`s from a data source. As part of this, `IOptions` was added to `SelectIdentityProvider`, which makes the signature the same as `GetIdentityProvider `.

`GetClaims` was modified to take a context argument. The reason it's not defaulted to null is because that would make it easy to mistakenly not provide that parameter in the library itself. The cost is that existing tests now have an extra null argument.

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/HttpRequestExtensions.cs
+++ b/Sustainsys.Saml2.AspNetCore2/HttpRequestExtensions.cs
@@ -10,7 +10,7 @@ namespace Sustainsys.Saml2.AspNetCore2
 {
     static class HttpRequestExtensions
     {
-        public static HttpRequestData ToHttpRequestData(
+        public static HttpRequestData<HttpContext> ToHttpRequestData(
             this HttpContext httpContext,
             ICookieManager cookieManager,
             Func<byte[], byte[]> cookieDecryptor)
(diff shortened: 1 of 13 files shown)


AndersAbel:
This is an interesting approach, thanks for submitting.

Since you are targetting the `develop` branch, there is no need to keep backwards compatibility. The `develop` will eventually turn into a version 3 release so no is the time to fix everything that requires breaking changes.

toddlucas:
This is one way to do it. From your original issue it appears that you want to make the base class itself take a generic parameter. This is essentially a bridge to that plan. When you're able to remove the Framework dependency (with StubIdp), the refactor should be trivial.

AndersAbel:
After thinking some more about future architecture this will not be needed. I will move forward with a direct dependency on the Asp.NET Core `HttpContext` and `HttpRequest`.

PR #1318, opened 2021-11-08
Improve logout information

vilhelmhei:
I have improved the logout handling slightly.  I first improved the logging to be able to trace what is going on. Then I extended the information that is sent to the listener of the logout notification.

1. First of all, when a custom logger is set in the clients code like this:

       authenticationBuilder.AddSaml2(authenticationScheme, options =>
       {
           options.SPOptions = new SPOptions
          {
              ...
              Logger = (ILoggerAdapter)logger
          };
       };
That logger got replaced by the default logger, so I introduced a check in for a custom logger before the default logger is set. That happens in the `PostConfigureSaml2Options.cs`

2. I added some verbose logging in `LogOutCommand.cs`.

3. The information sent in the `options.Notifications.LogoutCommandResultCreated` was very limited, only `Location` was set, so I added some details about the context of the logout to the `Content` member. I also adjusted the Unit Test to account for this change.

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/PostConfigureSaml2Options.cs
+++ b/Sustainsys.Saml2.AspNetCore2/PostConfigureSaml2Options.cs
@@ -23,7 +23,7 @@ namespace Sustainsys.Saml2.AspNetCore2
         /// <summary>
         /// Ctor
         /// </summary>
-        /// <param name="loggerFactory">Logger factory to use to hook up Saml2 loggin.</param>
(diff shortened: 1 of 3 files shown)


AndersAbel:
These are multiple changes in one PR and cannot be merged.
1. The PostConfigure bug fix looks correct, please submit in a separate PR.
2. Logging can always be improved, but the messages should be a bit more verbose to be readable without looking at the source code.
3. No, the Content field is not a place to put stuff. It is meant to be written as body content to the response and we do not want that with the received data.

PR #1321, opened 2021-11-16
Parse LogoutResponse XML values into Saml2LogoutResponse

henrikottesorensen:
In order to comply with the Danish NemLog-in logging policy, one has to log a long list of SAML2 properties. I’ve been able to get most log properties by hooking into the notifications at strategically picked points. However, I ran into a problem with logout responses, seems like SustainSys.Saml2 only parsed the status value from a LogoutResponse. I wanted the ID and InResponseTo attributes as well. 

Seemed straight forward enough to add deserialisation of the remaining attributes to Saml2LogoutResponse.FromXml().


milkshakeuk:
@AndersAbel any news on the v2.9.0 release?

PR #1329, opened 2021-12-19
Support .NET 6

Wael3bbas:
Fix Unable to cast object of type 'System.Security.Cryptography.RSACng' to type 'System.Security.Cryptography.RSACryptoServiceProvider "Sustainsys.Saml2" for .NET 6.0

### DIFF
--- a/Sustainsys.Saml2/Internal/EnvironmentHelpers.cs
+++ b/Sustainsys.Saml2/Internal/EnvironmentHelpers.cs
@@ -6,7 +6,7 @@ namespace Sustainsys.Saml2.Internal
 	public class EnvironmentHelpers
     {
         public static bool IsNetCore =>
-#if NETSTANDARD2_0
+#if NETSTANDARD2_0 || NET5_0 || NET6_0
         // To support .NET 5.0 and up, specifically check if this is not .NET Framework.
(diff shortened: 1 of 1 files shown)


AndersAbel:
Did you also change the project file to target .NET 5 or 6 when bulding to require this? If the build target is .Net Standard 2.0, this shouldn't be required as far as I now?

Wael3bbas:
My project is currently targeting `net6.0 `as below image and what I understand from [this Microsoft doc](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-5-and-net-standard) that `net5.0` replaces both `netcoreapp `and `netstandard `TFMs

image

AndersAbel:
Yes, that's correct. But if you build the library with the `netstandard2.0` moniker this shouldn't be needed.

What I'm trying to understand here is if this change is needed for a published nuget to be able to work with .NET 6 projects.

Wael3bbas:
In my case when I used the latest NuGet v2.8.0 it throw mentioned error and after downloading the code and doing this change it works fine with me.

I think it will be needed for both .NET 5 and 6 as `#if NETSTANDARD2_0` will always return `false`.

AndersAbel:
Ok, then this is not needed. v2.8.0 does not properly check for .NET5/6. v2.9.0 was just pushed to Nuget, it works with .NET 5 and 6.

PR #1330, opened 2021-12-20
added ability to customize metadata doc before signing

alexander-lysenko-88:
added ability to customize metadata xml document before signing it.
This will be handy for example in case when you need to add a node which is not supported by sustainsys out of the box.


AndersAbel:
Welcome as a contributor.

While I like the idea, there are a few things that need to be updated:
1. Please put whitespace fixes / using cleanups etc. in a separate commit - or even better - a separate PR. It's much harder to review functionality if they are mixed with the actual functionality changes.
2. I just released 2.9.0 today and after that I updated the branch guidelines (I assume you started work on this before). But from now on, I will not accept any more updates to `v2`, only to `develop` for new functionality. Please retarget your PR.

alexander-lysenko-88:
Ok, thanks for the feedback. I'll process the comments :)

alexander-lysenko-88:
@AndersAbel as I can see on the wiki the `develop` branch will be used for the `v3` release which will be Asp.Net Core compatible only. Does this mean that .Net Framework version is abandoned from now on? What if I'd like to use the functionality from .Net Framework App?

alexander-lysenko-88:
Something went wrong with my merge, so I will close this PR and create a new one

PR #1332, opened 2021-12-21
added ability to customize metadata xml document before signing it.

alexander-lysenko-88:
This will be handy for example in case when you need to add a node which is not supported by sustainsys out of the box.

NOTE: I've created a similar PR yesterday, but something went wrong with my merge, thus I've removed the old one and have created this one.


AndersAbel:
This functionality will be covered in `v3/develop`. Please see the new Metadata handling that is much more flexible than the old.

PR #1333, opened 2022-01-12
Read IgnoreAuthenticationContextInResponse from config

briandesarmo:
Allow `IgnoreAuthenticationContextInResponse` to be read from the config.


AndersAbel:
All of the web.config support is planned to be dropped in v3, so this is not relevant to merge.

If you need this functionality in a released package, it can be done as commercial work in a 2.X release. Please reach out to support@sustainsys.com if that's interesting.

PR #1339, opened 2022-04-10
Optional token replay validation

maw136:
This provides an option to disable Token Replay validation. In some integrations it is required, the defaults are kept unchanged.


AndersAbel:
This is against `v2` and no new PRs will be merged to `v2` unless security critical or part of paid support.

PR #1340, opened 2022-04-10
Added options for Saml2Urls customization

maw136:
Added an option to customize Command Urls. Some integrations require those to be different than default ones.


AndersAbel:
This is against `v2` and no new PRs will be merged to `v2` unless security critical or part of paid support.

PR #1373, opened 2022-10-26
Added Saml2Subject to AuthnRequest

### LINKED ISSUE #430: Subject/NameId in AuthnRequest
I have a scenario where i need support for specifying Subject/NameId in the AuthnRequest. The NameId must be "per login". I am thinking of using an URL query parameter.

I have tried this with success in a rough proof-of-concept but are now looking into adding this in a nice way following the projects structure, naming-convention etc. so that it can be followed by a PR.

Any tips, aside from the general guidelines for contributing as how to best implement this?

In my rough proof-of-concept i have the SignInCommand getting an instance of a (new) RequestNameId class from the HttpRequestData using a query string. The Saml2AuthenticationRequest then have an AddNameId method similar to the current AddNameIdPolicy.

I am also thinking of having the parsing of Subject/NameId enabled through configuration.

Any ideas are welcome. Thanks!

dhobi:
Should resolve #430
- Configurable either by URL or by relayData (Owin)

Please let me know if this is the correct approach or if I'd need to add other options as well. This just works for my use case where I want to pass a SAML "login_hint" to Keycloak.

If everything looks good, I'll add some tests as well.

### DIFF
--- a/Sustainsys.Saml2/SAML2P/Saml2AuthenticationRequest.cs
+++ b/Sustainsys.Saml2/SAML2P/Saml2AuthenticationRequest.cs
@@ -10,6 +10,7 @@ using System.Xml.Linq;
 using Sustainsys.Saml2.Configuration;
 using Sustainsys.Saml2.Internal;
 using Sustainsys.Saml2.WebSso;
+using Microsoft.IdentityModel.Tokens.Saml2;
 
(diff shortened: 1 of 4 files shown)


AndersAbel:
While this a feature that is appreciated, no new feature releases will be done to v2 (as stated on the readme), all efforts go into v3 (which is still at an early stage).

PR #1387, opened 2023-02-22
PLT-18499 Change ValidAudience check to be idp specific

uipathswapnil:
Prior to making this call to "CreateClaims", if the idp that is returned by the GetIdentityProvider notification has a different configuration then the default spOptions, then we see the following error:
image

This PR changes the validation audience based on the current identity provider rather then the entityid configured in spOptions. For our fork of the sustainsys library, this provided more customization as we have different identity providers per entityid. This change will allow users to check the audience against that specific identity provider.

### DIFF
--- a/Sustainsys.Saml2/IdentityProvider.cs
+++ b/Sustainsys.Saml2/IdentityProvider.cs
@@ -40,7 +40,7 @@ namespace Sustainsys.Saml2
             OutboundSigningAlgorithm = spOptions.OutboundSigningAlgorithm;
         }
 
-        readonly SPOptions spOptions;
+        internal readonly SPOptions spOptions;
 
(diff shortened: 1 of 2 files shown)


uipathswapnil:
@AndersAbel Please take a look and let me know if I can provide more information.

AndersAbel:
I'm sorry but In my opinion this is not a good behaviour. If you have one logical SP with multiple IdentityProviders, then the SP Entity Id is the same for all. If you have multiple logical SPs in your application with different Entity IDs, you should register the Saml2 module/handler/scheme multiple times, one for each logical SP.

To avoid spending time on creating a PR for functionality that I will not approve, please open an issue for discussion first.

Also - please note the information in the readme. New features will only be merged to `v2` for paying customers.

PR #1403, opened 2023-08-14
Explicitly target dotnet6 and use FrameworkReference in that case

Arjan321:
The `Sustainsys.Saml2.AspNetCore2` project references nuget packages that are no longer supported in `aspnetcore >=3`. 

This PR adds a new TFM (i picked `6`, but any version would do) and when using this new TFM, it uses [`FrameworkReference`](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/target-aspnetcore?view=aspnetcore-7.0&tabs=visual-studio#use-the-aspnet-core-shared-framework) instead of `PackageReference` to reference `aspnetcore` libraries


AndersAbel:
Thanks, this is a small change that makes sense

Arjan321:
Thanks! When can we expect a release?

AndersAbel:
Within a couple of weeks.

PR #1411, opened 2023-10-11
Use an IOptionsMonitor in the authentication handler

guillaume86:
Use an IOptionsMonitor in place of manually calling IOptionsFactory + IOptionsMonitorCache.

The change allows consumers to inject a custom IOptionsMonitor to implement for example multi tenancy. 
The official handlers all follow that pattern.

Here's the function in the default options monitor to reassure that this don't change existing behavior:
https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options/src/OptionsMonitor.cs#L90

### DIFF
--- a/Sustainsys.Saml2.AspNetCore2/Saml2Handler.cs
+++ b/Sustainsys.Saml2.AspNetCore2/Saml2Handler.cs
@@ -18,34 +18,30 @@ namespace Sustainsys.Saml2.AspNetCore2
     /// </summary>
     public class Saml2Handler : IAuthenticationRequestHandler, IAuthenticationSignOutHandler
     {
-        private readonly IOptionsMonitorCache<Saml2Options> optionsCache;
+        private readonly IOptionsMonitor<Saml2Options> optionsMonitor;
         private readonly IDataProtectionProvider dataProtectorProvider;
 
         // Internal to be visible to tests.
         internal Saml2Options options;
         HttpContext context;
         private IDataProtector dataProtector;
-        private readonly IOptionsFactory<Saml2Options> optionsFactory;
         bool emitSameSiteNone;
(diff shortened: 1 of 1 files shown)


AndersAbel:
I agree that following the standard pattern would be better - that's what I do in [`v3`](https://github.com/Sustainsys/Saml2/blob/develop/src/Sustainsys.Saml2.AspNetCore/Saml2Handler.cs).

But as I am focusing on `v3`, I try to not do any non-security fixes to `v1` or `v2`. To persuade me, can you tell me why you woul need a custom `IOptionsMonitor` for multi tenancy?

guillaume86:
Hey, sure I understand. 
I need to be able to change the options depending on a value in the HttpRequest, with a custom IOptionsMonitor it is pretty clean and easy to do and make it work with the caching by appending the tenant name to the cache key. But with the current implementation I don't have access to the cache key so it's basically impossible to implement. 
Have a nice day, thanks for the great library!

AndersAbel:
But I still don't understand why you need to customize the options for multi tenancy. The common approaches are to either register a scheme per tenant, or to have one scheme with multiple IdentityProviders in the collection?

guillaume86:
Your solutions only work when you have a static list of tenants at startup AFAIK. It's not my case I can add a new tenant without restart and a tenant can change its SSO settings at runtime.

AndersAbel:
Look at the `SelectIdentityProvider` and `GetIdentityProvider` notifications. If you implement both of those, you can read the Idp information from the DB.

If you read metadata for any of your Idps, you should cache the `IdentityProvider` objects in memory.

AndersAbel:
Can we close this as won't fix then?

PR #1449, opened 2024-04-06
Add testing for fetching attribute values

SwapnilShahGit:
This PR adds test cases for reading attributes in a SAML document. Alongside, there is a null check added in the code such that an attribute will not be populated if no value is specified.


src/Sustainsys.Saml2/Serialization/SamlXmlReader.Attribute.cs
```diff
@@ -40,8 +40,11 @@ protected virtual void ReadElements(XmlTraverser source, SamlAttribute attribute
         while(source.MoveNext(true) 
             && source.EnsureName(Elements.AttributeValue, Namespaces.SamlUri))
         {
-            // TODO: Test + support for null values.
-            attribute.Values.Add(source.GetTextContents());
+            var attributeValue = source.GetTextContents();
+            if (!string.IsNullOrEmpty(attributeValue))
+            {
+                attribute.Values.Add(attributeValue);
+            }
```
SwapnilShahGit, line 47: I am wondering if I need to add a check outside of the while loop to see if Values is populated. If it is not, then we should return null instead of a SamlAttribute object with a populated "Name" but an empty "Values".

AndersAbel:
Thanks for the PR. Unfortunately the SAML treatment of attribute values is a bit more complicated. The base part of the spec is section 2.7.3.1.1 in the SAML core spec, which includes how null and empty values are handled.

The TLDR (although I do recommend reading the spec):
* An empty node `<AttributeValue />` represents an empty string.
* A `null` value is represented by `<AttributeValue xsi:nil=\"1\" />` (`1` or `true` are both allowed)

The SAML Attribute Profiles (section 8, profile spec) contains even more rules, but as the main focus right now is the Web SSO Profile (and possibly Single Logout) I think we can ignore the profile spec for now.

src/Sustainsys.Saml2/Xml/XmlTraverser.cs
```diff
@@ -282,10 +282,17 @@ public bool EnsureName(string localName, string namespaceUri)
     /// Ensures that the contents of the current node is only text and returns the text.
     /// </summary>
     /// <returns></returns>
-    public string GetTextContents()
+    public string? GetTextContents()
     {
         //TODO: Test case + error handling for non text content.
         IgnoreChildren();
+        
+        var nullNodeValue = CurrentNode?.Attributes?.GetNamedItem("xsi:nil")?.Value;
```
AndersAbel, line 290: The support for null is specific to attribute values. The [Saml Core Spec](https://groups.oasis-open.org/higherlogic/ws/public/download/56776/sstc-saml-core-errata-2.0-wd-07.pdf/latest) section 2.7.3.1.1 states that:

> If a SAML attribute includes an empty value, such as the empty string, the corresponding
<AttributeValue> element MUST be empty (generally this is serialized as <AttributeValue/>).
This overrides the requirement in Section 1.3.1 that string values in SAML content contain at least one
non-whitespace character.
If a SAML attribute includes a "null" value, the corresponding <AttributeValue> element MUST be
empty and MUST contain the reserved xsi:nil XML attribute with a value of "true" or "1".

Checking 1.3.1 it states that string values must have at least one non-whitespace character.

I think that the correct validation to do in this method is to check that there is indeed at least one character, possibly with an "bool allowEmpty" param. Then the check for xsi:nil can be moved to `ReadElements(XmlTraverser source, SamlAttribute attribute)`.

And btw, the retrieval of the attribute should use the namespace. The prefix `xsi` is just a common thing to do, it doesn't have to be exactly that prefix.

`source.CurrentNode!.Attributes!["nil", XmlSchema.InstanceNamespace]?.Value`

AndersAbel (changes requested):

PR #1512, opened 2025-08-06
Add net 9.0 support

leorg99:
This adds .net 9.0 support. When packing, add both .net 8 and .net 9 binaries to the nuget package.


AndersAbel:
Thank for you submitting!

PR #1513, opened 2025-08-28
Validation

AndersAbel:


src/Sustainsys.Saml2/Validation/AssertionValidator.cs
```diff
@@ -14,8 +14,15 @@ public void Validate(
     {
         ValidateIssuer(assertion, parameters);
         ValidateConditions(assertion.Conditions, parameters);
+        ValidateSubjectConfirmation(assertion, parameters);
+        ValidateSubjectConfirmationData(assertion, parameters);
```
AndersAbel, line 18: I think ValidateSubjectConfirmationData should be called from ValidateSubjectConfirmation as it is part of the SubjectConfirmation. We should go one step further and add a ValidateSubject that calls ValidateSubjectConfirmation that calls ValidateSubjectConfirmationData.

ValidateSubject should check that there is a Subject. It is mandatory in the WebSSO profile so we should require it. The methods should not accept Assertions but instead be
```
ValidateSubject(Subject subject)
ValidateSubjectConfirmation(SubjectConfirmation confirmation)
ValidateSubjectConfirmationData(SubjectConfirmationData confirmationData).
```

This will make null checking easier.

src/Sustainsys.Saml2/Validation/AssertionValidator.cs
```diff
@@ -14,8 +14,15 @@ public void Validate(
...
+        // TODO: Validate Subject/SubjectConfirmation
```
AndersAbel, line 19: Remove TODO comments

src/Sustainsys.Saml2/Validation/AssertionValidator.cs
```diff
@@ -84,4 +91,50 @@ protected virtual void ValidateConditions(
         // to care about proxy restrictions. They are only valid if the SP is a proxy that acts as an IDP
         // to other applications. We'll just ignore proxy restrictions for now.
     }
+    /// <summary>
+    /// Validate Subject Confirmation of an assertion. 
+    /// </summary>
+    /// <param name="assertion"></param>
```
AndersAbel, line 97: Fill in XML docs descriptions of parameters. Can be one word each, but should at least be something.

src/Sustainsys.Saml2/Validation/AssertionValidator.cs
```diff
@@ -84,4 +91,50 @@ protected virtual void ValidateConditions(
         // to care about proxy restrictions. They are only valid if the SP is a proxy that acts as an IDP
         // to other applications. We'll just ignore proxy restrictions for now.
     }
+    /// <summary>
```
AndersAbel, line 94: Insert blank line before the /// for a new method (same in more places in this file)

src/Sustainsys.Saml2/Validation/AssertionValidator.cs
```diff
@@ -84,4 +91,50 @@ protected virtual void ValidateConditions(
...
+    /// Validate Subject Confirmation of an assertion. 
+    /// </summary>
+    /// <param name="assertion"></param>
+    /// <param name="parameters"></param>
+    /// <exception cref="SamlValidationException"></exception>
+    protected virtual void ValidateSubjectConfirmation(Assertion assertion, AssertionValidationParameters parameters)
+    {
+        var method = assertion.Subject.SubjectConfirmation!.Method;
```
AndersAbel, line 102: Add check (including test) for missing SubjectConfirmation.

Issue #1592, opened 2026-04-20
Add support for Sha256-rsa-MGF1


AndersAbel:
Ensure support for `http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1`

@glatzert Opening an issue here for the signing algorithm. BundID is a platform we want to support, so we'll add this natively.

glatzert:
Hi @AndersAbel, I'm willing to implement things like that myself and provide PRs, but I was a little bit unsure about where to look into the code itself - is v2 is the currently active developed version or should I look into main for that?

AndersAbel:
I think an additional algorithm can go into v2 and then we just lift it to v3. But it depends on your time frame, v3 is getting close to a first release.

glatzert:
I found two places where it was necessary to be added, and I got a successful authenticatin by adding it there.

PR #1593, opened 2026-04-20
Enable http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1

glatzert:
Allow `http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1` signature algorithm


glatzert:
Fixes #1592

romanett (approved):
I´d apreciate this to be merged

PR #1594, opened 2026-04-29
Enable EntitiesDescriptor to be read from metadata

glatzert:
In my mission to have something that's compatible with BundID, I started implementing the neccessary bits in v3 also.


glatzert:
Since I'm not an XML expert in any means, I could not provide the test case for the signed metadata. Perhaps, someone want's to pick that up?

